PR #15, opened 2021-08-12
Update rules wiki page

U8NWXD:


DubeySandeep:
@vojtechjelinek Can you please review these wiki page changes?

vojtechjelinek (approved):
LGTM!

Creating-Rules.md
```diff
@@ -1,51 +1,74 @@
...
-If you have just [created a new interaction](Creating-Interactions), say `MyInteraction` and now want to define the rules associated with it:
-  1. In [extensions/interactions/rules.json](https://github.com/oppia/oppia/blob/develop/extensions/interactions/rules.json) add a `MyInteraction` dictionary.
-  2. For each rule you want to create add an appropriate key (this is the “rule name”). The corresponding value should contain a single key, `description`, whose value is described in the Rule Description section.
-  3. Add a corresponding evaluation function, as described in the [Evaluation Function](#evaluation-function) section.
-  4. Add tests for the evaluation function, as described in the [Adding Tests](#adding-tests) section.
-
-Note that, if you simply want to add a new rule for a pre-existing interaction, say `MyInteraction`, then you only need to add a new key to the existing dictionary as in (2).
-
-### Rule Description ###
+## Table of contents
```
U8NWXD, line 3: Move to top

DubeySandeep:

Creating-Rules.md
U8NWXD: done

Creating-Rules.md
```diff
@@ -1,51 +1,74 @@
...
 ```
 
-The evaluation function may not use any other external parameters (such as the reader's current card, and so on) to compute its result; everything must be passed in through the subject or be defined when the rule is first initialized.
+Note that the evaluation function must not use any external parameters, for example the reader's current card. The `inputs` and `answer` must be sufficient to determine whether the rule is satisfied.
 
-## Adding Tests ##
+## Adding a New Rule
 
-You will need to create a `MyInteractionRulesServiceSpec.js` file containing tests for rules defined in the `myInteractionRulesService` factory service. Create a separate test for each rule and check the various corner cases to make sure that rules are defined correctly. You are strongly encouraged to do this for any new rule that you create, so that you can ensure that the rule works properly and returns the correct results on a variety of inputs.
+Suppose you have just [[created a new interaction|Creating-Interactions]], say `MyInteraction`, and now you want to define the rules associated with it. You should:
```
seanlip, line 67: You should: --> To do so, follow these steps:

(Otherwise, it reads as "You should ... in 'extensions/..., add a dictionary" etc., which is a little weird.)

Creating-Rules.md
```diff
@@ -1,51 +1,74 @@
...
 
-## Contributing New Rules ##
+1. In `extensions/interactions/rule_templates.json`, add a dictionary under the key `MyInteraction`. You can skip this step if you are adding a rule to an existing interaction.
+2. For each rule you want to create add an appropriate key that names the rule. The corresponding value should be a dictionary with a single key, `description`, whose value is a string describing the rule. The string should contain parameters as substrings of the form `{{paramName|ParamType}}`. See the [descriptions section above](#rule-descriptions) for details.
```
seanlip, line 70: Unfortunately 'param'/'parameters' is a term of art that is used for something else in Oppia (see http://oppia.github.io/#/Parameters). Although it's being deprecated, references to params still exist in the codebase. 

I think we typically call these "rule inputs", would it be possible to use that terminology instead?

seanlip (approved):

Creating-Rules.md
U8NWXD: done

Creating-Rules.md
U8NWXD: done

U8NWXD:
@seanlip I've addressed your comments PTAL

seanlip (approved):
Thanks, looks great!

PR #17, opened 2021-08-18
Update E2E tests page

U8NWXD:


End-to-End-Tests.md
```diff
@@ -1,228 +1,333 @@
...
+    * E2E tests are relatively "expensive," meaning that they take a while to run. Therefore, you should avoid testing something twice wherever possible. This usually means that fewer, larger tests are preferable to more, smaller tests. For example, consider these tests:
+
+      * Test exploration creation by creating a new exploration.
+      * Test exploration deletion by creating a new exploration and then deleting it.
+
+      Notice that we create an exploration in both tests. It would be more efficient to combine these into a single test:
+
+      * Test exploration creation and deletion by creating an exploration and then deleting it.
+
+2. Write the [utilities](#writing-utilities) you will need. Your test file should never interact with the page directly. All interactions should go through utilities. A good way to check that you're doing all page interactions through the utilities is to ensure that you have no element selectors (e.g. `element(by.css(...))`) in your suite files.
```
U8NWXD, line 207:
```suggestion
2. Write the [utilities](#writing-utilities) you will need. Your test file should never interact with the page directly. Use utilities instead. A good way to check that you're doing all page interactions through the utilities is to ensure that you have no element selectors (e.g. `element(by.css(...))`) in your suite files.
```

End-to-End-Tests.md
```diff
@@ -1,228 +1,333 @@
...
 
-Much of the difficulty of writing protractor code lies in specifying the element with which you wish to interact. It is important to do so in a way that is as insensitive as possible to superficial DOM features such as text and styling, so as to reduce the likelihood that changes will have to be made when the production html is changed. The order in which to attempt to specify an element in Oppia is as follows:
+2. Existing element ids, for example `explorationLanguageCode`. We avoid using existing classes for this purpose as they are generally style specifications such as `big-button` that may be changed in the future.
```
U8NWXD, line 221:
```suggestion
2. Using existing element ids, for example `explorationLanguageCode`. We avoid using existing classes for this purpose as they are generally style specifications such as `big-button` that may be changed in the future.
```

End-to-End-Tests.md
```diff
@@ -1,228 +1,333 @@
...
-  3. You can use `by.tagName` if you are sure you are in a context where only one element will have (or is likely to have in future) the given name. The `<input>` and `<button>` tags often fall under this category. Try to avoid `by.buttonText` and `by.linkText` since they are sensitive to the choice of user-facing text.
-  4. Finally, you can use `by.xpath` to specify an exact path from the starting element to the one you get to. This is not ideal since it renders the tests fragile to changes in the DOM. However sometimes it is necessary, for example to send elements to `expectRichText` in `forms.js` which requires it receive an exact element that cannot be directly specified.
+3. You can use `by.tagName` if you are sure you are in a context where only one element will have (or is likely to have in future) the given name. The `<input>` and `<button>` tags often fall under this category. Try to avoid `by.buttonText` and `by.linkText` since they are sensitive to the choice of user-facing text.
```
U8NWXD, line 223:
```suggestion
3. You can use `by.tagName` if you are sure you are in a context where only one element will have (and is likely to have in future) the given name. The `<input>` and `<button>` tags often fall under this category. Try to avoid `by.buttonText` and `by.linkText` since they are sensitive to the choice of user-facing text.
```

End-to-End-Tests.md
```diff
@@ -1,228 +1,333 @@
...
 
-Sometimes you need to distinguish between several different sibling elements, for which purpose you can use commands of the form
+4. Finally, you can use `by.xpath` to specify an exact path from the starting element to the one you get to. This is not ideal since it renders the tests fragile to changes in the DOM. However sometimes it is necessary, for example to send elements to `expectRichText` in `forms.js` which requires it receive an exact element that cannot be directly specified.
```
U8NWXD, line 225:
```suggestion
4. Finally, you can use `by.xpath` to specify an exact path from the starting element to the one you get to. This is not ideal since it renders the tests fragile to changes in the DOM.
```

End-to-End-Tests.md
```diff
@@ -1,228 +1,333 @@
...
+
+```text
+
+       Root
+       /  \
+      /   ...
+     /      \
+         Element A: class="protractor-test-elem-a"
+             \
+    ...      ...
```
U8NWXD, line 237: Fix left `...`

End-to-End-Tests.md
```diff
@@ -1,228 +1,333 @@
...
+      await action.click('Topic button ${i}', button);
+      return True;
+    }
+  }
+  return False;
+}
+```
 
-It can happen that a test fails because the webpage has not finished loading functionality by the point at which protractor attempts to use it; failures of this kind can be inconsistent between successive runs of the same test. If one occurs you can try the following:
+It might be tempting to use the `.first()`, `.last()`, and `.get(n)` functions directly when you know what order the elements will come in. However, this makes the tests fragile to changes in the page, and it makes the code hard to read. You should also avoid accessing page elements by index since that's not how most users will find elements. They will be relying on the text identifying those elements, and your test should too.
```
U8NWXD, line 275:
```suggestion
It might be tempting to use the `.first()`, `.last()`, and `.get(n)` functions directly when you know what order the elements will come in. However, this makes the tests fragile to changes in the page, and it makes the code hard to read. You should also avoid accessing page elements by index because that's not how most users will find elements. They will be relying on the text identifying those elements, and your test should too.
```

End-to-End-Tests.md
```diff
@@ -1,228 +1,333 @@
...
 
 * HTML tags should be unique if possible. When they are not unique, for instance when multiple copies of the same HTML are created dynamically, we should not find one with indexing, `.first()`, or `.last()`. A great example of how to do this correctly is in `this.playTutorial` in `ExplorationEditorMainTab.js`.
-    * There is really only one case where it is acceptable to identify HTML elements by index, which is when the following conditions all hold:
-        1. The elements you want to choose among are siblings, meaning that they share the same parent element. If they aren't siblings, then you can add use the parent elements to distinguish between them, for example by adding HTML classes to the parents.
-        2. The elements you want to choose among are identical. In particular, if the elements contain different text, then you can use that text to distinguish them.
-        3. The elements you want to choose among are generated dynamically, so you can't modify them to add HTML classes.
+
+  * There is really only one case where it is acceptable to identify HTML elements by index, which is when the following conditions all hold:
```
U8NWXD, line 291:
```suggestion
  There is really only one case where it is acceptable to identify HTML elements by index, which is when the following conditions all hold:
```

End-to-End-Tests.md
```diff
@@ -1,228 +1,333 @@
...
+    3. The elements you want to choose among are generated dynamically, so you can't modify them to add HTML classes.
+
 * Avoid for loops where the loop index is used in asynchronous calls. `this.expectHintsTabContentsToMatch` in `ExplorationEditorTranslationTab.js` is a better way because it puts the index in the CSS selector, so the index is used before the asynchronous part kicks in.
-* Do not use URLs to:
-    * Get IDs, for example IDs for explorations, collections, or topics
-    * Navigate to a page, for example opening the about page by navigating to `/about` instead of clicking the appropriate buttons
+
+* Do not use URLs to navigate to a page, for example opening the about page by navigating to `/about` instead of clicking the appropriate buttons
+
 * Do not use `browser.sleep(` calls. This is great for debugging, but in the final test you should use `waitFor` instead.
```
U8NWXD, line 301:
```suggestion
* Do not use `browser.sleep()` calls. They are fine for debugging, but in the final test you should use `waitFor` instead.
```

End-to-End-Tests.md
```diff
@@ -259,23 +365,35 @@ Whenever you're debugging tests, you should create a debugging doc to document y
    await adminPage.viewRolesbyUsername('moderator1');
    ...
    ```
-2. Run the e2e script with the flag --debug_mode.
-   eg. `python -m scripts.run_e2e_tests --debug_mode --suite="topicAndStoryEditor"`
-3. Wait for the script to show the following log:
+
+2. Run the e2e script with the flag --debug_mode. For example,
```
U8NWXD, line 369:
```suggestion
2. Run the e2e script with the flag `--debug_mode`. For example,
```

End-to-End-Tests.md
```diff
@@ -259,23 +365,35 @@ Whenever you're debugging tests, you should create a debugging doc to document y
...
+
+   ![Inspect Page](https://user-images.githubusercontent.com/11008603/88563290-714bac80-d04f-11ea-8b36-fc43c66d6e3d.png)
+
 6. A Chrome dev tools instance will open up and the e2e test should start executing in a new window.
+
 7. The control will stop at the point where the debugger statement was added. You can now choose to inspect elements, log variables in the test, or add more break-points.
 
 ### Rerunning with SSH
+
 Circle CI allows debugging using SSH. For details, please read [this](https://circleci.com/docs/2.0/ssh-access-jobs/#steps). Debugging with SSH only reruns that particular job, so it is a great way to rerun a passing test to see if it flakes without rerunning all the tests.
```
U8NWXD, line 394: Circle CI -> CircleCI

ditto elsewhere

End-to-End-Tests.md
```diff
@@ -319,52 +437,89 @@ If you don’t see the `Artifacts` link, go to the summary of the failing workfl
...
 
 ## Metrics
 
-We track passes, known flakes, and failures that aren't known to be flakes (called "failures") using a logging server. You can view these metrics at https://oppia-e2e-test-results-logger.herokuapp.com.
+We track passes, known flakes, and failures that aren't known to be flakes (called "failures") using a logging server. You can view these metrics at https://oppia-e2e-test-results-logger.herokuapp.com. The code for this server is in the (private) [`oppia/oppia-e2e-test-results-logger`](https://github.com/oppia/oppia-e2e-test-restuls-logger) repository.
 
 ## Reference
 
-### Forms and objects
+### Forms and Objects
```
U8NWXD, line 448:
```suggestion
### Forms and objects
```

End-to-End-Tests.md
```diff
@@ -509,4 +664,37 @@ which will work for both the editing and checking cases.
   var output = await someAsynchronousFunction();
```
U8NWXD, line 664: Above, add languages to code blocks in async-await tips

End-to-End-Tests.md
```diff
@@ -509,4 +664,37 @@ which will work for both the editing and checking cases.
...
   await // do something with "output"
```
U8NWXD, line 665: Above, no `i` in `.first()` and `.last()`

U8NWXD:
@DubeySandeep @seanlip PTAL

End-to-End-Tests.md
```diff
@@ -1,228 +1,344 @@
...
+It might be tempting to use the `.first()`, `.last()`, and `.get(n)` functions directly when you know what order the elements will come in. However, this makes the tests fragile to changes in the page, and it makes the code hard to read. You should also avoid accessing page elements by index because that's not how most users will find elements. They will be relying on the text identifying those elements, and your test should too.
 
-  1. Use ExpectedConditions to determine if the element in question is either present, visible or clickable, depending on your use case.
-  2. If you are attempting to use an element that is not visible you can try `general.scrollElementIntoView(elem)` to get to it, though normally this should happen automatically.
+Except for cases where an element selector is crafted dynamically, all element selectors should be at the top of a utility file. There should be no element selectors in suite files.
```
seanlip, line 288: I just realized that "utility file" is not defined, and there is no mention of pageobjects anywhere. This seems like an omission? The separation of concerns seems important to understanding where to put what, and why we lay things out that way.

End-to-End-Tests.md
```diff
@@ -299,17 +428,17 @@ then you can look for the report that includes this message. Times like `10003ms
 
 Sometimes you'll get screenshots that just aren't very helpful. For example, a lot of screenshots show the login page for some reason. You can check other examples of a flake though. One of the others might have a useful screenshot.
 
-### Videos of End-to-end Tests
+### Downloading screen recordings
 
-The Automated QA team has implemented a video recording system -- now, videos of end-to-end test runs on GitHub Actions may be accessed. This system will help developers solve problems in e2e tests that only occur on CI or are difficult to replicate locally.
+When screen recordings are enabled, we capture video of the test running on GitHub Actions. This helps developers solve problems in E2E tests that only occur on CI or are difficult to replicate locally.
```
seanlip, line 433: Link this to an explanation for how to enable?

End-to-End-Tests.md
```diff
@@ -319,52 +448,89 @@ If you don’t see the `Artifacts` link, go to the summary of the failing workfl
 
 ![Artifacts section in summary of e2e run](https://user-images.githubusercontent.com/52176783/118647358-502c9600-b7f2-11eb-9e41-a6f5f962ddfb.png)
 
-People on macOS -- Quicktime doesn’t seem to like the videos we generate, so you might need to use VLC media player to view the videos.
+Note for macOS: Quicktime doesn’t seem to like the videos we generate, so you might need to use VLC media player to view the videos.
 
 ## Metrics
 
-We track passes, known flakes, and failures that aren't known to be flakes (called "failures") using a logging server. You can view these metrics at https://oppia-e2e-test-results-logger.herokuapp.com.
+We track passes, known flakes, and failures that aren't known to be flakes (called "failures") using a logging server. You can view these metrics at https://oppia-e2e-test-results-logger.herokuapp.com. The code for this server is in the (private) [`oppia/oppia-e2e-test-results-logger`](https://github.com/oppia/oppia-e2e-test-restuls-logger) repository.
```
seanlip, line 455: Is it useful to add the last sentence? If so should we explain why it's private?

I lean slightly towards omitting it since it's not actually that useful, but your call.

seanlip (changes requested):
Looks nice! Took a pass :)

End-to-End-Tests.md
U8NWXD: removed

End-to-End-Tests.md
U8NWXD: I clarified the later paragraph that describes when recordings are generated

End-to-End-Tests.md
U8NWXD: done

End-to-End-Tests.md
```diff
@@ -1,228 +1,344 @@
...
 
-### If You Are Adding End-to-End Tests
+--------------------- time ---------------------->
+```
 
-First, please make sure the tests aren't flaky by running them on your own a bunch of times. You could do this by running them locally or on your own TravisCI fork.
+The durations of steps `A` and `B` are non-deterministic because `A` depends on how quickly the browser executes the frontend code to open the modal, and `B` depends on how fast the test code runs. Since these operations are happening on separate processes, the operating system makes no guarantees about which will complete first. In other words, we have a race condition.
 
-Next, add your tests to a separate test suite. On TravisCI, this should run as its own job. On CircleCI, add this as another `run` command to an existing job. This will make it easy for us to tell whether a failure is due to your test or not.
+This race condition means that the test can fail randomly even when there's nothing wrong with Oppia's code. These failures are called _flakes_.
```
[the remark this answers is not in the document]
U8NWXD: done

U8NWXD:
@seanlip I've addressed your comments. PTAL

End-to-End-Tests.md
```diff
@@ -436,7 +440,7 @@ Each individual test within each suite gets its own video. The video of each tes
 
 ![Name of video for test gets printed out above test](https://user-images.githubusercontent.com/52176783/118647333-486cf180-b7f2-11eb-999b-9edbbb89b5a7.png)
 
-Note that no videos will be generated on CircleCI due to memory issues when running the video recorder. Further, videos are disabled on GitHub Actions where the `VIDEO_RECORDING_IS_ENABLED` environment variable is set to `0`.
+Note that while recordings are enabled by default, no videos will be generated on CircleCI due to memory issues when running the video recorder. Further, videos are disabled on GitHub Actions where the `VIDEO_RECORDING_IS_ENABLED` environment variable is set to `0`.
```
seanlip, line 443: >> Link this to an explanation for how to enable?
> 
> I clarified the later paragraph that describes when recordings are generated

It seems that it is 0 for everything at the moment, so I'm not sure this counts as "recordings enabled by default".

seanlip (changes requested):

End-to-End-Tests.md
U8NWXD: By "default" I meant that recordings will be saved so long as the environment variable is not set to 0. However, that's not actually correct--the environment variable has to be set to 1 for videos to be saved. I've updated the wiki page to reflect that

U8NWXD:
@seanlip I've made fixes based on your comments. PTAL

End-to-End-Tests.md
seanlip: Yup, but what I'm trying to point out is that the main thing developers are going to look for when reading this section is "how do I get screen recordings for my currently-failing test". I think it might be worth having that really clear and front-and-center, rather than just a description of what the system currently is.

More generally, you might want a step-by-step list somewhere of "these are the steps you should take when debugging an e2e test" (one of which presumably involves getting screen recordings).

U8NWXD: Ahh, that makes sense. done

U8NWXD:
@seanlip PTAL

seanlip (approved):
Looks much better. Thanks!

PR #18, opened 2021-08-19
Update frontend tests page

U8NWXD:


Frontend-unit-tests-guide.md
```diff
@@ -1,409 +1,484 @@
...
 
-### Handling with asynchronous code  
+### Handling asynchronous code
+
+#### Making HTTP calls
 
-#### Making HTTP calls  
-In the frontend, while writing tests, we don’t make actual calls to the backend. All HTTP calls are mocked since the frontend tests actually run without a backend in place.  
-Similarly, any services can also be mocked. We try to keep the usage of such mocks as low as possible since the more mocks there are, the more divergence there is with the underlying code.  
+In the frontend, while writing tests, we don’t make actual calls to the backend. All HTTP calls are mocked since the frontend tests actually run without a backend available. Similarly, any services can also be mocked. We try to keep the usage of such mocks as low as possible since the more mocks there are, the more divergence there is with the underlying code.
```
U8NWXD, line 436:
```suggestion
All HTTP calls must be mocked since the frontend tests actually run without a backend available. Similarly, any services can also be mocked. We try to keep the usage of such mocks as low as possible since the more mocks there are, the more divergence there is between the mocks and the code being mocked.
```

Frontend-unit-tests-guide.md
```diff
@@ -1,409 +1,484 @@
...
-```
+
+In order to make HTTP calls in a secure way, it's common that applications have tokens to authenticate the user while they are using the platform. In the codebase, there is a specific service to handle the token, called CsrfTokenService. When mocking HTTP calls, you must mock this service in the test file so the tests won't fail (you don't have a real token when running unit tests so it needs to be mocked). Then, you should just copy and paste this piece of code inside a beforeEach block (the CsrfService will be a variable with the return of `$injector.get('CsrfTokenService')` -- in AngularJS -- or `TestBed.get(CsrfTokenService)` -- in Angular 2+):
```
U8NWXD, line 440:
```suggestion
In order to make HTTP calls in a secure way, it's common that applications have tokens to authenticate the user while they are using the platform. In the codebase, there is a specific service to handle the token, called CsrfTokenService. When mocking HTTP calls, you must mock this service in the test file so the tests won't fail due to lacking a token. Then, you should just copy and paste this piece of code inside a beforeEach block (the CsrfService will be a variable with the return of `$injector.get('CsrfTokenService')` -- in AngularJS -- or `TestBed.get(CsrfTokenService)` -- in Angular 2+):
```

Frontend-unit-tests-guide.md
```diff
@@ -1,409 +1,484 @@
...
-##### AngularJS
-[Here](https://github.com/oppia/oppia/blob/ae649aa08f1375457ec9e3c90257197b68fec7cd/core/templates/domain/learner_dashboard/learner-playlist.service.spec.ts#L84-L99) is an example which uses `$httpBackend` to mock the backend responses. A brief version of the code there, with some explanatory comments is given below.     
-To mock a backend call, you need to use `$httpBackend` dependency. There are two ways to expect a HTTP method (you can use both):
-- `$httpBackend.expectMETHODNAME(URL)` - like `expectPOST` or `expectGET` for instance
-- `$httpBackend.expect(‘METHOD’, URL)` - You pass the HTTP method as the first argument.  
+##### HTTP calls in AngularJS
+
+[Here](https://github.com/oppia/oppia/blob/ae649aa08f1375457ec9e3c90257197b68fec7cd/core/templates/domain/learner_dashboard/learner-playlist.service.spec.ts#L84-L99) is an example which uses `$httpBackend` to mock the backend responses.
+
+To mock a backend call, you need to use `$httpBackend` dependency. There are two ways to expect an HTTP method (you can use both):
```
U8NWXD, line 454:
```suggestion
To mock a backend call, you need to use the `$httpBackend` dependency. There are two ways to expect an HTTP method (you can use both):
```

Frontend-unit-tests-guide.md
```diff
@@ -1,409 +1,484 @@
...
+Just like the AngularJS way to mock HTTP calls, the Angular 2+ has flush functions to return the expected response and execute the mock correctly.
 
-#### Using `done` and `done.fail` from Jasmine  
-Using `done` and `done.fail` is another way to test asynchronous code in Jasmine. You can use it on promises (HTTP calls and so on), timers as `setTimeout`.  
+#### Using `done` and `done.fail` from Jasmine
 
-There’s a specific case where you should use `done` on mocking HTTP calls: when you want to assert the result of the fulfilled or reject promise, as you can see [here](https://github.com/oppia/oppia/blob/2e60d69d7b/core/templates/services/assets-backend-api.service.spec.ts#L274-L292). In this piece of code, we need to assert the response variable, then we use `done` after doing the assertion so Jasmine understands the asynchronous code has been completed. You can use `done.fail` when handling with rejected promises.
+Using `done` and `done.fail` is another way to test asynchronous code in Jasmine. You can use it on promises (HTTP calls and so on) and timers as `setTimeout`.
```
U8NWXD, line 469:
```suggestion
Using `done` and `done.fail` is another way to test asynchronous code in Jasmine. You can use it on promises (HTTP calls and so on) and timers such as `setTimeout`.
```

Frontend-unit-tests-guide.md
```diff
@@ -413,59 +488,56 @@ it('should wait for 10 seconds to call console.log', function() {
...
+#### Mocking with `$q` API in AngularJS
 
-### When should the upgraded services be imported in the test file?  
-One of the active projects in Oppia is the Angular2+ migration. By now, the AngularJS services are being migrated and it’s still being used in downgrade files. When testing AngularJS files which uses an Angular2+ as a dependency, you must use the beforeEach call below: 
-Let us assume that the test requires MyExampleService which is an Angular2+ service.
+When mocking a promise in AngularJS, you might use the `$q` API. In these cases, you must use `$scope.$apply()` or `$scope.$digest` because it forces `$q` promises to be resolved through a Javascript digest. Here are some examples using [$apply](https://github.com/oppia/oppia/blob/2e60d69d7b/core/templates/pages/email-dashboard-pages/email-dashboard-page.controller.spec.ts#L101-L108) and [$digest](https://github.com/oppia/oppia/blob/2e60d69d7b/core/templates/pages/exploration-editor-page/services/exploration-states.service.spec.ts#L209-L221).
```
U8NWXD, line 502:
```suggestion
When mocking a promise in AngularJS, you might use the `$q` API. In these cases, you must use `$scope.$apply()` or `$scope.$digest` because they force `$q` promises to be resolved through a Javascript digest. Here are some examples using [$apply](https://github.com/oppia/oppia/blob/2e60d69d7b/core/templates/pages/email-dashboard-pages/email-dashboard-page.controller.spec.ts#L101-L108) and [$digest](https://github.com/oppia/oppia/blob/2e60d69d7b/core/templates/pages/exploration-editor-page/services/exploration-states.service.spec.ts#L209-L221).
```

Frontend-unit-tests-guide.md
```diff
@@ -413,59 +488,56 @@ it('should wait for 10 seconds to call console.log', function() {
...
 
-So, what you need to do is to change the order of beforeEach calls, as you can see in this [test](https://github.com/oppia/oppia/blob/2e60d69d7b/core/templates/pages/exploration-editor-page/improvements-tab/services/improvement-suggestion-thread-modal.controller.spec.ts#L36-L51).
+### beforeEach calls in AngularJS
 
-### How to handle common errors  
-- If you see an error like `Error: Trying to get the Angular injector before bootstrapping the corresponding Angular module`, it means you are using a service (directly or indirectly) that is Upgraded to Angular and this error can throw or two reasons:
-  - Your test that is written in AngularJS is unable to get that particular service.
-    You can fix this by providing the value of Angular2+ service using $provide.
+If you’re testing an AngularJS file that uses an upgraded service, you’ll need to include a beforeEach block which mocks all the upgraded services. [Here](https://github.com/oppia/oppia/blob/2e60d69d7b/core/templates/pages/exploration-editor-page/editor-tab/training-panel/training-modal.controller.spec.ts#L35-L48)) is an example.
```
U8NWXD, line 527:
```suggestion
If you’re testing an AngularJS file that uses an upgraded service, you’ll need to include a beforeEach block which mocks all the upgraded services. [Here](https://github.com/oppia/oppia/blob/2e60d69d7b/core/templates/pages/exploration-editor-page/editor-tab/training-panel/training-modal.controller.spec.ts#L35-L48) is an example.
```

Frontend-unit-tests-guide.md
```diff
@@ -413,59 +488,56 @@ it('should wait for 10 seconds to call console.log', function() {
...
 
-    Let us assume that the test requires MyExampleService which is an Angular2+ service.
-    ```
+However, you might face the following situation: you need to mock an Angular2+ service by using `$provide.value`. Here’s the problem: if you use `$provide.value` before calling the updated services, your mock will be overwritten by the original code of the service. So, you need to change the order of beforeEach calls, as you can see in [this test](https://github.com/oppia/oppia/blob/2e60d69d7b/core/templates/pages/exploration-editor-page/improvements-tab/services/improvement-suggestion-thread-modal.controller.spec.ts#L36-L51).
+
+### How to handle common errors
+
+* If you see an error like `Error: Trying to get the Angular injector before bootstrapping the corresponding Angular module`, it means you are using a service (directly or indirectly) that is upgraded to Angular. This error can be thrown for two reasons:
```
U8NWXD, line 533: We don't have two reasons

U8NWXD:
@DubeySandeep @seanlip PTAL

U8NWXD:
Also @nithusha21 could you take a look at these changes?

Frontend-unit-tests-guide.md
```diff
@@ -1,409 +1,484 @@
...
+* `absoluteValue(0)` and `absoluteValue(1)`: These test cases are not comprehensive because they do not test negative numbers, and it's important for an absolute value function to correctly handle negative inputs. However, the code coverage is 100% because both blocks of the `if` statement are executed.
+* `absoluteValue(-1)`, `absoluteValue(0)`, and `absoluteValue(1)`: These test cases are comprehensive, and code coverage is 100%. Note that even though line 1 doesn't execute, coverage is 100% because line 1 is not executable.
+
+This example illustrates something very important about code coverage: **Code coverage less than 100% implies that the tests are not comprehensive, but code coverage of 100% does not imply that tests are comprehensive.** Therefore, while code coverage is a useful tool, you should primarily think about whether your tests cover all the possible behaviors of the code being tested--you should have a behavior-first perspective. Don't just think about which lines are covered.
```
seanlip, line 94: Perhaps replace "--" with "In other words,"; at the moment it's not entirely clear what "--" implies and I had to read it twice.

Also, possibly the part to bold should be "whether your tests cover all the possible behaviors of the code being tested". That seems like the key thing to emphasize here.

Frontend-unit-tests-guide.md
```diff
@@ -1,409 +1,484 @@
...
+6     return number
+7   }
+8 }
+```
+
+Now consider the following sets of test cases:
+
+* `absoluteValue(1)` and `absoluteValue(5)`: These test cases are not comprehensive because they only test positive numbers. Code coverage is also not 100% because line 4 is never executed.
+* `absoluteValue(0)` and `absoluteValue(1)`: These test cases are not comprehensive because they do not test negative numbers, and it's important for an absolute value function to correctly handle negative inputs. However, the code coverage is 100% because both blocks of the `if` statement are executed.
+* `absoluteValue(-1)`, `absoluteValue(0)`, and `absoluteValue(1)`: These test cases are comprehensive, and code coverage is 100%. Note that even though line 1 doesn't execute, coverage is 100% because line 1 is not executable.
```
seanlip, line 92: This is a fantastic explanation and a very nice example!

Frontend-unit-tests-guide.md
```diff
@@ -1,409 +1,484 @@
...
 A unit test is made of functions that configure the test environment, make assertions, and separate the different contexts of each situation. There are some test functions that are used across the codebase:
-- #### **describe**  
-  The describe function has a string parameter which should contain the name of the component being tested or     (when nested within another describe function) should describe the conditions imposed on the specific context pertaining to the tests in that “describe” block. Here are some examples:
-  ```
+
+#### describe
+
+The describe function has a string parameter which should contain the name of the component being tested or (when nested within another describe function) should describe the conditions imposed on the specific context pertaining to the tests in that “describe” block. Here are some examples:
+
+* An ouer describe function:
```
seanlip, line 154: Typo: ouer

Frontend-unit-tests-guide.md
```diff
@@ -1,409 +1,484 @@
...
 ### Unit test structure
+
 A unit test is made of functions that configure the test environment, make assertions, and separate the different contexts of each situation. There are some test functions that are used across the codebase:
-- #### **describe**  
-  The describe function has a string parameter which should contain the name of the component being tested or     (when nested within another describe function) should describe the conditions imposed on the specific context pertaining to the tests in that “describe” block. Here are some examples:
-  ```
+
+#### describe
+
+The describe function has a string parameter which should contain the name of the component being tested or (when nested within another describe function) should describe the conditions imposed on the specific context pertaining to the tests in that “describe” block. Here are some examples:
```
seanlip, line 152: Should "describe" be written as `describe` throughout (where its literal English meaning would otherwise not make sense in context)? Similarly for other code keywords.

Frontend-unit-tests-guide.md
```diff
@@ -1,409 +1,484 @@
...
+#### Validate external side-effects
 
-If you have any questions about the above, you can contact any of the people below:
-* @aishwary023 - aishwary.saxena.min19@iitbhu.ac.in
-* @gp201 - praneethg2001@gmail.com
-* @Radesh-kumar - imradesh@gmail.com
+An external side-effect is an effect that we expect from running the function but that isn't part of the outputs. For example, consider this pseudocode:
+
+```text
+function logIn(username, password) {
```
seanlip, line 351: I think we use `login` in general, right? Ditto below in setLoginCookie.

Frontend-unit-tests-guide.md
```diff
@@ -1,409 +1,484 @@
...
+  valid = whether the username and password are valid
+  if valid {
+    setLogInCookie(username)
+    return True
+  }
+  return False
+}
+```
 
+Here, the call to `setSessionCookie()` is an external side-effect. In our unit tests, it's not enough to check that the function's return value is correct; we also need to check that `setLogInCookie()` is called (and not called) correctly.
```
seanlip, line 361: What is setSessionCookie here?

Frontend-unit-tests-guide.md
```diff
@@ -1,409 +1,484 @@
...
+
+#### Spying on third-party libraries
 
-#### Spying on and handling with third-party libraries  
 Also, the spy can be used when mocking third-party libraries, like JQuery, mostly when doing ajax calls. [Here](https://github.com/oppia/oppia/blob/2e60d69d7b/core/templates/services/assets-backend-api.service.spec.ts#L274-L292)'s a good example when mocking JQuery ajax calls.
 
-#### Spying the same method/property more than one time in same context  
-It is impossible to spy twice the same method or property in the same context (block). For instance, the code below would throw an error:
-```
+#### Spying on the same method/property more than one time in same context
```
seanlip, line 385: Missing "the" before "same"

Frontend-unit-tests-guide.md
```diff
@@ -1,409 +1,484 @@
...
+
+It is impossible to spy twice on the same method or property in the same context (block). For instance, the code below would throw an error:
```
seanlip, line 387: Maybe explain what "context (block)" is -- I'm not sure what it means.

Frontend-unit-tests-guide.md
```diff
@@ -476,63 +548,75 @@ So, what you need to do is to change the order of beforeEach calls, as you can s
...
+* `*.service.ts`
+* `*Factory.ts`
+* `*.factory.ts`
+* `*.tokenizer.ts`
+
+As a good first issue, all the services that need to be tested are listed in [issue #4057](https://github.com/oppia/oppia/issues/4057).
+
+### Testing AngularJS services
+
+Use these files that are correctly following the testing patterns as reference:
```
seanlip, line 567: Here and below -- should it be "for reference" or "as references"?

seanlip (changes requested):
Thanks @U8NWXD! I took a pass.

Frontend-unit-tests-guide.md
U8NWXD: yep! done

Frontend-unit-tests-guide.md
U8NWXD: I've changed this to "scope"

Frontend-unit-tests-guide.md
U8NWXD: done

Frontend-unit-tests-guide.md
U8NWXD: that's a mistake! fixed

Frontend-unit-tests-guide.md
U8NWXD: done

Frontend-unit-tests-guide.md
U8NWXD: done

Frontend-unit-tests-guide.md
U8NWXD: done

Frontend-unit-tests-guide.md
U8NWXD: done

U8NWXD:
@seanlip I've addressed your comments. PTAL

seanlip (approved):
Thanks! LGTM.

U8NWXD:
@DubeySandeep PTAL

nithusha21 (approved):

PR #19, opened 2021-08-25
Update objects page

U8NWXD:


DubeySandeep:
@vojtechjelinek Thanks for the review!

@seanlip Can you please take the final pass?

Creating-Objects.md
```diff
@@ -1,12 +1,105 @@
...
+}
+```
+
+The schema would match a dictionary like this:
+
+```json
+{"version": 34}
+```
+
+The `dict` type is a collection, but the `int` type is an object. We define these objects as classes in [`extensions/objects/models/objects.py`](https://github.com/oppia/oppia/blob/develop/extensions/objects/models/objects.py) and components in [`extensions.objects/templates`](https://github.com/oppia/oppia/blob/develop/extensions/objects/templates).
```
seanlip, line 34: I don't understand the first sentence here. What is a "collection" in this context?

Creating-Objects.md
```diff
@@ -1,12 +1,105 @@
...
+            dict. The object schema.
+        """
+        return {
+            'type': 'int'
+        }
+```
+
+Object classes all meet the following criteria:
+
+* Inherit from either `BaseObject` or another object class.
```
seanlip, line 63: "They inherit...", "Their `description` class variable is set to...", etc.

That way, you describe them in terms of what they are (as suggested by the above intro to the list) and not in terms of what the developer does.

Creating-Objects.md
```diff
@@ -1,12 +1,105 @@
...
+* Set the class variable `default_value` to the value that should be the default whenever this type is used in a rule. Note that this variable is technically not required if the object will never be used in a [[rule|Creating-rules]].
+* Provide a `normalize` class method that accepts a raw Python object. The method should first validate the raw object to check whether it is malformed. If the validation checks pass, then the method should return a normalized form of the object. Note that in this context, "normalized" means a primitive Python type or multiple primitive types combined with lists and/or dicts.
+
+  * The `BaseObject` class provides a `normalize` class method that can handle normalization for you so long as your object class provides a `get_schema` class method that returns a schema describing the object. `BaseObject.normalize` passes your schema to `schema_utils.normalize_against_schema`, so make sure that `normalize_against_schema` won't try to normalize your object to your schema using your object class. Otherwise, you'll get an endless loop between `normalize_against_schema` and `BaseObject.normalize`.
```
seanlip, line 68: Probably not "your object"? The "object class" seems reasonable as "your" since the reader is presumably writing that, but the objects that come into the normalize() method are external ones over which the reader has no control.

Creating-Objects.md
```diff
@@ -1,12 +1,105 @@
...
+}
+```
+
+The `schema-based-editor` component already knows how to construct a form for this schema.
+
+## Create a new object
+
+Let's suppose you're creating a new object called `MyObject` (in practice you should use a more descriptive name). You'll need to follow these steps:
+
+1. Create a new [object class](#python-classes) named `MyObject`.
```
seanlip, line 98: Maybe state that it should satisfy the properties you mentioned above?

Creating-Objects.md
```diff
@@ -1,12 +1,105 @@
...
+2. Add tests for your class in `extensions/objects/models/objects_test.py`. You should add normalization tests to the `ObjectNormalizationUnitTests` class. You can also create a new test class to hold additional test cases.
+3. Add a new [Angular component](#angular-components) to accept input in the form of your object. Your component files should be named `my-object-editor.component.*`, and your component class should be named `MyObjectEditorComponent`.
+4. Import your component in `extensions/objects/object-components.module.ts` and add it to the module there.
+
+## Using objects
+
+You are unlikely to use objects directly. Instead, you will reference them as you [[create interactions|Creating-Interactions]], [[define rules|Creating-Rules]], or otherwise use [[schemas|Schemas]].
```
seanlip, line 105: Not really sure what the alternative of "use objects directly" means -- aren't the latter cases examples of "using objects directly"?

seanlip (changes requested):

Creating-Objects.md
U8NWXD: The "collection" here refers to the collections mentioned in the previous paragraph:

> These schemas can describe arbitrarily nested collections of base types. We call these base types _objects_.

I've re-worked the sentence to focus on the objects, since that's the important part.

Creating-Objects.md
U8NWXD: done

Creating-Objects.md
U8NWXD: done

Creating-Objects.md
U8NWXD: done

Creating-Objects.md
U8NWXD: I removed that bit

U8NWXD:
@seanlip PTAL

Creating-Objects.md
```diff
@@ -31,7 +31,7 @@ The schema would match a dictionary like this:
 {"version": 34}
 ```
 
-The `dict` type is a collection, but the `int` type is an object. We define these objects as classes in [`extensions/objects/models/objects.py`](https://github.com/oppia/oppia/blob/develop/extensions/objects/models/objects.py) and components in [`extensions.objects/templates`](https://github.com/oppia/oppia/blob/develop/extensions/objects/templates).
+This schema describes a dictionary of ints, and the `int` type is an object. We define these objects as classes in [`extensions/objects/models/objects.py`](https://github.com/oppia/oppia/blob/develop/extensions/objects/models/objects.py) and components in [`extensions.objects/templates`](https://github.com/oppia/oppia/blob/develop/extensions/objects/templates).
```
seanlip, line 34: Doesn't this describe a dict of one key-value pair where the key is 'version' and the value is an int? This has a different sense than "a dictionary of ints".

seanlip (changes requested):

Creating-Objects.md
U8NWXD: yep! fixed

U8NWXD:
@seanlip PTAL

Creating-Objects.md
```diff
@@ -31,7 +31,7 @@ The schema would match a dictionary like this:
 {"version": 34}
 ```
 
-This schema describes a dictionary of ints, and the `int` type is an object. We define these objects as classes in [`extensions/objects/models/objects.py`](https://github.com/oppia/oppia/blob/develop/extensions/objects/models/objects.py) and components in [`extensions.objects/templates`](https://github.com/oppia/oppia/blob/develop/extensions/objects/templates).
+This schema describes a dictionary with a `version` key whose value is an `int`, and the `int` type is an object. We define these objects as classes in [`extensions/objects/models/objects.py`](https://github.com/oppia/oppia/blob/develop/extensions/objects/models/objects.py) and components in [`extensions.objects/templates`](https://github.com/oppia/oppia/blob/develop/extensions/objects/templates).
```
seanlip, line 34: Sorry, this is still a bit confusing. You say "whose value is an int, and the int type is an object" -- this makes it unclear whether int is a type, value, or object.

Actually on re-reading this I think you might be confusing two things. Firstly, there are standard python types: int, bool, etc. -- these are not objects defined by us. 

Secondly, there are custom types like Int, UnicodeString, etc. These are types defined by us and is what is contained in objects.py. Note that objects are not necessarily base types, so the statement "We call these base types 'objects'" is wrong -- though they typically have an internal structure that is constructed out of base Python types. For example, the Int object in objects.py has type "int", and more complex objects have more complex types. Objects are invoked when the 'type' in the schema dict is 'custom'.

Does this clarify? Happy to chat and explain more about this if you like.

seanlip (changes requested):

Creating-Objects.md
U8NWXD: Yes, I think your explanation helped. I took another stab at explaining what objects are.

U8NWXD:
@seanlip PTAL

Creating-Objects.md
```diff
@@ -9,16 +9,16 @@
...
 ```json
 {
   "type": "dict",
   "properties": [
     {
-      "name": "version",
+      "name": "length",
       "schema": {
-        "type": "int"
+        "type": "Real"
```
seanlip, line 21: I think this is still wrong. Schemas that use an object have type='custom' and an obj_type key. See e,g, state_domain_test.py line 4339 and the schema parsing logic in schema_utils.py

Creating-Objects.md
```diff
@@ -9,16 +9,16 @@
 
 ## Introduction
 
-When collecting user input, we often define the kind of input we expect using [[schemas|Schemas]]. These schemas can describe arbitrarily nested collections of base types. We call these base types _objects_. For example, consider this schema:
+When collecting user input, we often define the kind of input we expect using [[schemas|Schemas]]. These schemas can describe arbitrarily nested collections of simple types. We call these simple types _objects_. For example, consider this schema:
```
seanlip, line 12: I think this -- "These schemas can describe arbitrarily nested collections of simple types. We call these simple types _objects_" -- is still incorrect. 

An object is basically a custom type. It doesn't really matter whether it's arbitrary nested or simple or whatever ... some objects are simple and some aren't, some objects are nested and some aren't. It's just a custom type, typically with its own frontend viewer/editor component (see extensions/objects).

Creating-Objects.md
```diff
@@ -28,10 +28,12 @@ When collecting user input, we often define the kind of input we expect using [[
 The schema would match a dictionary like this:
 
 ```json
-{"version": 34}
+{"length": 34.5}
 ```
 
-This schema describes a dictionary with a `version` key whose value is an `int`, and the `int` type is an object. We define these objects as classes in [`extensions/objects/models/objects.py`](https://github.com/oppia/oppia/blob/develop/extensions/objects/models/objects.py) and components in [`extensions.objects/templates`](https://github.com/oppia/oppia/blob/develop/extensions/objects/templates).
+This schema describes a dictionary with a `length` key whose value is of type `Real`, and the `Real` type is one of our objects. We define these objects as classes in [`extensions/objects/models/objects.py`](https://github.com/oppia/oppia/blob/develop/extensions/objects/models/objects.py) and components in [`extensions.objects/templates`](https://github.com/oppia/oppia/blob/develop/extensions/objects/templates).
```
seanlip, line 34: Should extensions.objects be extensions/objects?

Creating-Objects.md
```diff
@@ -28,10 +28,12 @@ When collecting user input, we often define the kind of input we expect using [[
...
+
+Note that schemas can also use Python base types like `int`. These types are not objects, even though many of our objects are very similar to Python base types. For example, the `Real` object is essentially a `float`. However, `float` is not one of our objects.
```
seanlip, line 36: Schemas have their own definition for which keys are allowed for which types -- see schema_utils.py for details, or https://github.com/oppia/oppia/wiki/Schemas/8774d9fb027d82f39a6bbb80c8c08737615fa28d#write-a-schema

The 'type' key can only take a certain allowed set of values, not just any custom / builtin type.

seanlip (changes requested):

Creating-Objects.md
U8NWXD: fixed

Creating-Objects.md
U8NWXD: fixed

Creating-Objects.md
U8NWXD: done

Creating-Objects.md
U8NWXD: I removed this paragraph

U8NWXD:
@seanlip PTAL

seanlip (approved):
Thanks -- LGTM!

PR #29, opened 2021-09-15
Add frontend test debugging page

U8NWXD:


Debug-frontend-tests.md
```diff
@@ -0,0 +1,72 @@
...
+* [Isolating tests](#isolating-tests)
+* [Verbose mode](#verbose-mode)
+* [Finding stack elements](#finding-stack-elements)
+
+## Introduction
+
+Debugging frontend tests can be difficult because the stack traces are much less helpful than in the backend. There are two reasons for this:
+
+* Because so much of our frontend code is asynchronous, the stack trace will often just point to code that executes a queue of jobs when you usually want to know what code submitted the job.
+* Jasmine, our testing framework, does not support showing accurate line numbers in stack traces. We looked into fixing this in [#9648](https://github.com/oppia/oppia/issues/9648) but abandoned the idea once it was clear that Jasmine wasn't interested in supporting the feature.
```
U8NWXD, line 14: too harsh

Debug-frontend-tests.md
```diff
@@ -0,0 +1,72 @@
...
+	    at runInTestZone (core/templates/combined-tests.spec.js:427336:38)
+	    at UserContext.<anonymous> (core/templates/combined-tests.spec.js:427351:24)
+Chrome Headless 83.0.4103.116 (Mac OS 10.15.5): Executed 1 of 2277 (1 FAILED) (0 secs / 0.046 secs)
+```
+
+The `zone.js` stack elements relate to the [zone.js](https://github.com/angular/angular/tree/master/packages/zone.js) package that is handling our asynchronous operations. Notice that the line numbers are unreasonably large. This is because to run our tests, we load all our testing code and dependencies into `combined-tests.spec.js`. The line numbers refer to the resulting "mega-file," which doesn't actually exist on the file system.
+
+Below, we'll discuss some strategies you can use to overcome these difficulties.
+
+## Finding tests
```
U8NWXD, line 38: Make these headings present tense

U8NWXD:
@nithusha21 @srijanreddy98 could you please take a look at this PR. In particular, do you have any suggestions for other debugging strategies I could add?

Debug-frontend-tests.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Isolate tests
+
+Once you have identified which test is failing, you often want to run that test in isolation next. This will let you get results quickly as you try other debugging strategies. Thankfully, you just have to change the `it()` of your test to `fit()`. You can even do this multiple times to mark multiple tests for running. Then when you run the frontend tests, only those tests with `fit()` will run. Note that `fdescribe()` works the same way.
+
+**Remember to revert your `fit()` and `fdescribe()` changes before committing!**
+
+## Verbose mode
+
+By default, we suppress log output from the tests to the terminal when running the frontend tests. This keeps the test output clean so it's easy to find which tests failed. However, this output can be useful for debugging. For example, if you add `console.log()` or `console.error()` statements to your tests for debugging, the output from those statements will be suppressed by default. To view this output, pass `--verbose` when you run the frontend tests:
```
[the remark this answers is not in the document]
U8NWXD: This is just for debugging, so these statements shouldn't be committed. In that case, the linter doesn't matter, right?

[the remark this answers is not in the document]
gp201: During our, GSoC to debug or test the flake fix we used the following [code](https://gist.github.com/gp201/3bb7e649680b01eed8d3370c7e44e506). We then pushed this to our fork and created a PR to our fork so the GitHub actions would run on our fork.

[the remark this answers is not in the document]
U8NWXD: Okay. I'll add a note about the linter. I think `console.log` is actually useful because if you aren't testing on CI, you might _want_ the linter to remind you in case you forget to remove your debugging statements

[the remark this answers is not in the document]
U8NWXD: done

Debug-frontend-tests.md
```diff
@@ -0,0 +1,72 @@
+## Table of contents
+
+* [Introduction](#introduction)
+* [Find tests](#find-tests)
+* [Isolate tests](#isolate-tests)
+* [Verbose mode](#verbose-mode)
+* [Find stack elements](#find-stack-elements)
```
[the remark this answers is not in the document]
U8NWXD: done. I left the table of contents as-is since that's a standard pattern across the wiki pages, but I added steps in the introduction

U8NWXD:
@gp201 PTAL

Debug-frontend-tests.md
```diff
@@ -0,0 +1,98 @@
...
+    it('should handle rejection when fetching a file fails', () => {
+      ...
+    });
+  });
+});
+```
+
+The strings from the `describe` and `it` functions get concatenated with space (` `) delimiters to create the test name. The name of the `it` block usually starts with `should`, so we recommend searching the codebase for the portion of the test name from `should` to the end (e.g. "should handle rejection when fetching a file fails" in this case).
+
+## Isolate tests
```
gp201, line 59: FYI, sometimes certain tests flake only when the previous tests run in a certain pattern. For example:

```
describe('', () => {
var test = 2;
beforeEach(() => {
...
});

it('test 1', () => {
test = 3;
});

it('test 2', () => {
expect(test).toBe(2);
})
});
```

in the above case if `test-1` runs before `test-2` only then will it fail. This might be a useful tip for developers

Debug-frontend-tests.md
```diff
@@ -0,0 +1,98 @@
...
+* `console.error()` statements do not cause a lint failure, so they work well when you are pushing your code with debugging code to a PR to let the tests run in the CI environment.
+
+## Find stack elements
+
+While you can't use the line numbers in a stack trace to find the associated code, you can sometimes use the function or class names. For example, in the example above we could try searching the codebase for `_maybeConvertBody`. The success of this technique depends on how frequently the string you search for occurs in the codebase. If the string occurs a lot, you will have a hard time figuring out which occurrence is the one you are interested in.
+
+## Run tests repeatedly on CI
+
+Especially when debugging flaky frontend tests, you may want to set your tests to run many times on CI. For example, you might add some debugging code and then rerun your frontend tests on CI until the flake occurs. You can update the frontend test CI workflow to enable this. In `.github/workflows/frontend_tests.yml`, change the step that runs the tests to:
```
gp201, line 86: Do we recommend them to run the GitHub action on their own fork? Or is it okay to run on the Oppia repo itself?

gp201 (changes requested):

Debug-frontend-tests.md
U8NWXD: done

Debug-frontend-tests.md
U8NWXD: We want them to use their own fork. I've added that

U8NWXD:
@gp201 PTAL

gp201 (approved):
LGTM

U8NWXD:
@DubeySandeep PTAL

DubeySandeep (approved):

U8NWXD:
@seanlip PTAL

Debug-frontend-tests.md
```diff
@@ -0,0 +1,118 @@
...
+    test = 3;
+  });
+
+  it('should do something else', () => {
+    expect(test).toBe(2);
+  });
+});
+```
+
+These tests will fail if `should do something` runs before `should do something else`, but not if the tests run in the reverse order. Our frontend tests run in a non-deterministic order, so these kinds of problems generally appear as flakes. In other words, these tests will sometimes pass but fail other times, even if the underlying code is the same.
```
seanlip, line 81: We should perhaps talk about trying to avoid such cases and that it is our goal to not have that. If a developer finds such a case should they report it, fix it, etc.?

Debug-frontend-tests.md
```diff
@@ -0,0 +1,118 @@
...
+**Only use --verbose when you are running a few tests in isolation. Otherwise, you will be swamped with way too many log messages.**
+
+Note that depending on your situation, `console.log()` or `console.error()` might be preferable:
+
+* `console.log()` statements will cause the linter to fail, so they are a great choice for local debugging. Then if you forget to remove them before pushing your changes, the linter will remind you.
+* `console.error()` statements do not cause a lint failure, so they work well when you are pushing your code with debugging code to a PR to let the tests run in the CI environment.
+
+## Find stack elements
+
+While you can't use the line numbers in a stack trace to find the associated code, you can sometimes use the function or class names. For example, in the example above we could try searching the codebase for `_maybeConvertBody`. The success of this technique depends on how frequently the string you search for occurs in the codebase. If the string occurs a lot, you will have a hard time figuring out which occurrence is the one you are interested in.
```
seanlip, line 100: What should they do in the "hard time" case?

Debug-frontend-tests.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Run tests repeatedly on CI
+
```
seanlip, line 103: Does the term "flaky" need to be defined?

seanlip (changes requested):

Debug-frontend-tests.md
U8NWXD: I just removed it

Debug-frontend-tests.md
U8NWXD: They should pick a different stack element to search for. I've added text to explain this

Debug-frontend-tests.md
U8NWXD: I think we want developers to treat these cases as incorrect tests and fix them. I've added a note to state this

U8NWXD:
@seanlip PTAL

Debug-frontend-tests.md
```diff
@@ -78,7 +78,9 @@ describe('oppia', () => {
 });
 ```
 
-These tests will fail if `should do something` runs before `should do something else`, but not if the tests run in the reverse order. Our frontend tests run in a non-deterministic order, so these kinds of problems generally appear as flakes. In other words, these tests will sometimes pass but fail other times, even if the underlying code is the same.
+These tests will fail if `should do something` runs before `should do something else`, but not if the tests run in the reverse order. Our frontend tests run in a non-deterministic order, so these kinds of problems generally appear as intermittent failures. In other words, these tests will sometimes pass but fail other times, even if the underlying code is the same.
+
+The tests should work correctly regardless of what order they run in. If you find some tests where it matters what order the tests run in, those tests are incorrect, so you should fix them just like you would fix a failing test.
```
seanlip, line 83: Should we add a note (maybe both here + the frontend tests page) to please not write any tests that break this condition / please adhere to these condition when writing tests?

seanlip (approved):
Thanks, LGTM, just one additional comment but feel free to handle it as you see fit!

Debug-frontend-tests.md
U8NWXD: done

PR #30, opened 2021-09-15
Diagnose PR failures page

U8NWXD:


U8NWXD:
@DubeySandeep PTAL

DubeySandeep (approved):

U8NWXD:
@seanlip PTAL

_Sidebar.md
```diff
@@ -37,13 +37,13 @@
     * [[How to write type definitions|Guide-on-defining-types]]
     * [[How to write design docs|Writing-design-docs]]
     * [[Revert and Regression Policy|Revert-and-Regression-Policy]]
-    * [[If your build fails...|If-Your-Build-Fails]]
+    * Debugging
+      * [[Diagnose PR failures|Diagnose-PR-failures]]
```
seanlip, line 41: The title of this page seems a bit off because it's not the PR that's failing. Maybe something like "What to do if CI checks on PRs fail" or "What to do if there's a red cross next to your PR"?

_Sidebar.md
```diff
@@ -37,13 +37,13 @@
...
       * [[If your presubmit checks fail|If-your-presubmit-check-fails]]
       * [[Interpreting CircleCI Results | Interpreting CircleCI Results]]
-      * Debugging
-        * [[Debugging Docs|Debugging-Docs]]
-        * [[Debugging end-to-end tests|Debug-end-to-end-tests]]
-        * [[Debugging Stories|Debugging-Stories]]
+      * [[Debugging Docs|Debugging-Docs]]
```
seanlip, line 44: Have you thought through organizing this sublist structurally? It seems a bit confusing in its current form, e.g. "Debugging" is used differently in the fourth and sixth points compared to the fifth.

Maybe something like:

- Diagnosing presubmit check failures
- Diagnosing GitHub Actions failures  (note: is "GitHub Actions failures" the way a developer would refer to this? Or is it "CI checks"? But the former would include the next bullet point, and the latter would also include the next couple of bullet points...)
- Diagnosing end-to-end test failures
- Diagnosing CircleCI failures
- Writing debugging docs
- Debugging Examples

Maybe the last of these could be integrated in the relevant part of one of the first four bullet points? Also more generally, do we need to deduplicate some of this content (since the second bullet point seems a bit like a catch-all)?

seanlip (changes requested):
Thanks @U8NWXD! Mainly a general question about organization, I suspect more might need to be done here.

_Sidebar.md
U8NWXD: Changed to `If-CI-checks-fail-on-your-PR.md`

U8NWXD:
@seanlip PTAL

seanlip (approved):
Thanks! LGTM.

PR #31, opened 2021-09-16
Build process

U8NWXD:


Build-process.md
```diff
@@ -0,0 +1,115 @@
...
+    * [Maintenance mode](#maintenance-mode)
+    * [Deploy mode](#deploy-mode)
+    * [Unused: Deparallelized terser mode](#unused-deparallelized-terser-mode)
+* [Source maps](#source-maps)
+
+## Introduction
+
+Before we can run the Oppia server, we have to build it. This build process transforms the files defining our application (e.g. code, images, configuration files, etc.) from the developer-friendly format we store in GitHub into a form that can be executed to run the server. If you are running the local development server, this build process makes minimal changes, but when deploying to production, we do much more optimization.
+
+This page documents how the build process works in each of its modes. For most developers the build steps are handled automatically and can be ignored; however, some bugs appear only in some build modes. For example, a bug might only happen in prod mode, in which case a developer will be unable to reproduce it in a local development mode.
```
U8NWXD, line 20:
```suggestion
This page documents how the build process works in each of its modes. For most developers the build steps are handled automatically and can be ignored; however, some bugs appear only in certain build modes. For example, a bug might only happen in prod mode, in which case a developer will be unable to reproduce it in dev mode.
```

Build-process.md
```diff
@@ -0,0 +1,115 @@
...
+## Build modes
+
+There are two primary build modes: dev mode and prod mode. When running in prod mode, you can also enable maintenance mode or deploy mode, or you can choose to only minify third-party libraries. This means the build mode hierarchy looks like this:
+
+* Dev mode
+* Prod mode
+  * Maintenance mode
+  * Deploy mode
+  * Minify only third-party libraries
+We use prod mode for most everything besides local development. We use prod mode when deploying to production and testing servers, and we run most tests in prod mode. A notable exception is the lighthouse accessibility tests, which run in dev mode.
```
U8NWXD, line 35:
```suggestion
We use prod mode for most everything besides local development. For example, we use prod mode when deploying to production and testing servers, and we run most tests in prod mode. A notable exception is the lighthouse accessibility tests, which run in dev mode.
```

Build-process.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Build modes
+
+There are two primary build modes: dev mode and prod mode. When running in prod mode, you can also enable maintenance mode or deploy mode, or you can choose to only minify third-party libraries. This means the build mode hierarchy looks like this:
+
+* Dev mode
+* Prod mode
+  * Maintenance mode
+  * Deploy mode
+  * Minify only third-party libraries
```
U8NWXD, line 34:
```suggestion
  * Minify only third-party libraries

```

Build-process.md
```diff
@@ -0,0 +1,115 @@
...
+We use prod mode for most everything besides local development. We use prod mode when deploying to production and testing servers, and we run most tests in prod mode. A notable exception is the lighthouse accessibility tests, which run in dev mode.
+
+
+### Build mode constants
+
+The build mode is controlled by the following constants in `assets/constants.ts`:
```
U8NWXD, line 40:
```suggestion
The build mode affects the application during runtime through the following constants in `assets/constants.ts`:
```

Build-process.md
```diff
@@ -0,0 +1,115 @@
...
+
+* `DEV_MODE`: Set to `true` if and only if we are in dev mode. Otherwise, this is `false`.
+* `EMULATOR_MODE`: Set to `true` if and only if we are not in deploy mode. Note that being in deploy mode implies being in prod mode. Otherwise, this is `false`.
+* `ENABLE_MAINTENANCE_MODE`: Set to `true` if and only if we are in maintenance mode. Note that being in maintenance mode implies being in prod mode. Otherwise, this is `false`.
+
+### Dev mode
+
+We use dev mode for local development. For example, when you run `python -m scripts.start`, you build the app in dev mode. This mode is optimized to build as quickly as possible, so it doesn't do much optimization. This means that the server doesn't run quite as fast.
```
U8NWXD, line 48:
```suggestion
We use dev mode for local development. For example, when you run `python -m scripts.start`, you build the app in dev mode. This mode is optimized to build as quickly as possible, so it doesn't do much optimization. This means that the server doesn't run quite as quickly.
```

Build-process.md
```diff
@@ -0,0 +1,115 @@
...
+
+Here's what happens during a dev mode build:
+
+1. Third-party libraries are generated. If you recently ran the server, then there might not be much to change here. Otherwise, the third-party libraries from `manifest.json` are downloaded. The downloaded CSS and JavaScript code files are combined into `third_party/generated/third_party.css` and `third_party/generated/third_party.js`, and the downloaded fonts are placed in `third_party/webfonts/`.
+2. The build script sets the constants in `feconf.py` to enable dev mode. Specifically, `DEV_MODE` is `true`, and the other constants are `false`.
```
U8NWXD, line 53:
```suggestion
2. The build script sets the constants in `assets/constatns.ts` to enable dev mode. Specifically, `DEV_MODE` is `true`, and the other constants are `false`.
```

Build-process.md
```diff
@@ -0,0 +1,115 @@
...
+3. [Webpack](https://webpack.js.org/) bundles our frontend code files and runs with a [configuration](https://github.com/oppia/oppia/blob/develop/webpack.dev.config.ts) designed for minimal compilation time. Note that the dev mode configuration file inherits from the [root webpack config file](https://github.com/oppia/oppia/blob/develop/webpack.common.config.ts). In dev mode, webpack runs in "watch" mode so that when files are changed, webpack re-builds the app automatically. Note that **this step is not handled by `build.py`.** Instead, this step is usually performed by whatever script you actually executed, for example `start.py`. When running the frontend tests without minifying third-party libraries, this step doesn't happen at all.
```
U8NWXD, line 54:
```suggestion
3. [Webpack](https://webpack.js.org/) bundles our frontend code files and runs with a [configuration](https://github.com/oppia/oppia/blob/develop/webpack.dev.config.ts) designed for minimal compilation time. Note that the dev mode configuration file inherits from the [root webpack config file](https://github.com/oppia/oppia/blob/develop/webpack.common.config.ts). In dev mode, webpack runs in "watch" mode so that when files are changed, webpack re-builds the app automatically. Note that **this step is not handled by `build.py`.** Instead, this step is usually performed by whatever script you actually executed, for example `start.py`. When running the frontend tests, this step doesn't happen at all.
```

Build-process.md
```diff
@@ -0,0 +1,115 @@
...
+
+1. The third-party libraries are generated like in [dev mode](#dev-mode), but they are also minified so they take up less space.
+2. Hashes are generated for asset files like images. These will be used later.
+3. [Webpack](https://webpack.js.org) runs with watch mode disabled and a configuration optimized for application performance.
+4. `app.yaml` is generated from `app_dev.yaml`. The two files are identical in plain prod mode, except for a comment in `app.yaml` noting that it is auto-generated.
+5. The third-party libraries, assets, files compiled by webpack, and some other files are all copied to a build folder `build/`. This mimics the build folder we generate and upload to production servers when deploying the app for real. During this process, asset files are renamed to include the hashes we generated earlier. This ensures that if the file content changes, the name changes, so any caches of the old file are invalidated.
+
+This is the mode our tests run in, and it's what you'll want to use if you want to locally run a version of the app that's as close to the production version as possible. You can use this mode like this:
```
U8NWXD, line 71:
```suggestion
This is the mode our tests (except for the lighthouse accessibility tests) run in, and it's what you'll want to use if you want to locally run a version of the app that's as close to the production version as possible. You can use this mode like this:
```

Build-process.md
```diff
@@ -0,0 +1,115 @@
...
+* `EMULATOR_MODE`: Set to `true` if and only if we are not in deploy mode. Note that being in deploy mode implies being in prod mode. Otherwise, this is `false`.
+* `ENABLE_MAINTENANCE_MODE`: Set to `true` if and only if we are in maintenance mode. Note that being in maintenance mode implies being in prod mode. Otherwise, this is `false`.
+
+### Dev mode
+
+We use dev mode for local development. For example, when you run `python -m scripts.start`, you build the app in dev mode. This mode is optimized to build as quickly as possible, so it doesn't do much optimization. This means that the server doesn't run quite as fast.
+
+Here's what happens during a dev mode build:
+
+1. Third-party libraries are generated. If you recently ran the server, then there might not be much to change here. Otherwise, the third-party libraries from `manifest.json` are downloaded. The downloaded CSS and JavaScript code files are combined into `third_party/generated/third_party.css` and `third_party/generated/third_party.js`, and the downloaded fonts are placed in `third_party/webfonts/`.
```
U8NWXD, line 52: Note that this is only some of the third-party libs and all in the frontend

Build-process.md
```diff
@@ -0,0 +1,115 @@
...
+python -m scripts.start --prod_env
+```
+
+#### Minify only third-party libraries
+
+In this mode, the third-party libraries are installed and minified, but no other steps from [plain prod mode](#plain-prod-mode) happen. This is only used by the frontend tests.
+
+#### Maintenance mode
+
+As far as the build process is concerned, maintenance mode works just like [plain prod mode](#plain-prod-mode) except that the `ENABLE_MAINTENANCE_MODE` constant gets set to `true`. Then when the app runs, only admins can log in. We use this when we are upgrading the production server and need to let jobs run without letting users modify any data.
```
U8NWXD, line 83:
```suggestion
As far as the build process is concerned, maintenance mode works just like [plain prod mode](#plain-prod-mode) except that the `ENABLE_MAINTENANCE_MODE` constant gets set to `true`. Then when the app runs, only admins can log in. We use this when we are upgrading the production server and need to let jobs run while ensuring users don't change any data.
```

U8NWXD:
@DubeySandeep @vojtechjelinek PTAL

Build-process.md
```diff
@@ -0,0 +1,116 @@
...
+The build process is performed by [`scripts/build.py`](https://github.com/oppia/oppia/blob/develop/scripts/build.py). If you ever want to know exactly how a particular part of the build process works, you should check out the code there.
+
+## Build modes
+
+There are two primary build modes: dev mode and prod mode. When running in prod mode, you can also enable maintenance mode or deploy mode, or you can choose to only minify third-party libraries. This means the build mode hierarchy looks like this:
+
+* Dev mode
+* Prod mode
+  * Maintenance mode
+  * Deploy mode
```
vojtechjelinek, line 33: Just a note: both maintenance mode and prod mode can be enabled together, we usually expect the maintenance mode be enabled with deploy mode so it might be possible to include the maintenance mode under deploy mode.

Build-process.md
```diff
@@ -0,0 +1,116 @@
...
+  * Minify only third-party libraries
+
+We use prod mode for most everything besides local development. For example, we use prod mode when deploying to production and testing servers, and we run most tests in prod mode. A notable exception is the lighthouse accessibility tests, which run in dev mode because they don't depend on prod mode, and dev mode compilation is faster.
```
vojtechjelinek, line 36: I think we sometimes use prod model for local testing (in order to get "closer" to the deployed version)

Build-process.md
```diff
@@ -0,0 +1,116 @@
...
+
+The build mode affects the application during runtimes through the following constants in `assets/constants.ts`:
+
+* `DEV_MODE`: Set to `true` if and only if we are in dev mode. Otherwise, this is `false`.
+* `EMULATOR_MODE`: Set to `true` if and only if we are not in deploy mode. Note that being in deploy mode implies being in prod mode. Otherwise, this is `false`.
+* `ENABLE_MAINTENANCE_MODE`: Set to `true` if and only if we are in maintenance mode. Note that being in maintenance mode implies being in prod mode. Otherwise, this is `false`.
+
+### Dev mode
+
+We use dev mode for local development. For example, when you run `python -m scripts.start`, you build the app in dev mode. This mode is optimized to build as quickly as possible, so it doesn't do much optimization. This means that the server doesn't run quite as quickly.
```
vojtechjelinek, line 49: I would exclude the last sentence. The speed of dev × prod when ran locally is ± the same, this is because compression is not used locally (compression provides huge perf bonus).

Build-process.md
```diff
@@ -0,0 +1,116 @@
...
+
+Here's what happens during a dev mode build:
+
+1. Some third-party libraries are generated. If you recently ran the server, then there might not be much to change here. Otherwise, the third-party libraries from `manifest.json` are downloaded. Note that the libraries in `manifest.json` are just some of our frontend libraries. The downloaded CSS and JavaScript code files are combined into `third_party/generated/third_party.css` and `third_party/generated/third_party.js`, and the downloaded fonts are placed in `third_party/webfonts/`.
+2. The build script sets the constants in `assets/constants.ts` to enable dev mode. Specifically, `DEV_MODE` is `true`, and the other constants are `false`.
```
vojtechjelinek, line 54: emulator mode is `true` too

Build-process.md
```diff
@@ -0,0 +1,116 @@
...
+3. [Webpack](https://webpack.js.org/) bundles our frontend code files and runs with a [configuration](https://github.com/oppia/oppia/blob/develop/webpack.dev.config.ts) designed for minimal compilation time. Note that the dev mode configuration file inherits from the [root webpack config file](https://github.com/oppia/oppia/blob/develop/webpack.common.config.ts). In dev mode, webpack runs in "watch" mode so that when files are changed, webpack re-builds the app automatically. Note that **this step is not handled by `build.py`.** Instead, this step is usually performed by whatever script you actually executed, for example `start.py`. When running the frontend tests, this step doesn't happen at all.
```
vojtechjelinek, line 55:
```suggestion
3. [Webpack](https://webpack.js.org/) bundles our frontend code files and runs with a [dev configuration](https://github.com/oppia/oppia/blob/develop/webpack.dev.config.ts) designed for minimal compilation time. Note that the dev mode configuration file inherits from the [root webpack config file](https://github.com/oppia/oppia/blob/develop/webpack.common.config.ts). In dev mode, webpack runs in "watch" mode so that when files are changed, webpack re-builds the app automatically. Note that **this step is not handled by `build.py`.** Instead, this step is usually performed by whatever script you actually executed, for example `start.py`. When running the frontend tests, this step doesn't happen at all.
```

vojtechjelinek: Also frontend test also build webpack but it is done through configuration provided in karma.conf.ts

Build-process.md
```diff
@@ -0,0 +1,116 @@
...
+
+In prod mode, we more aggressively optimize the app for performance, but the cost of this optimization is that the build process takes longer. Prod mode is also a little more complicated because it has three additional options: minify only third-party libraries, maintenance mode, and deploy mode.
+
+#### Plain prod mode
+
+First, let's consider what happens in plain prod mode where those three options are all disabled:
+
+1. The third-party libraries are generated like in [dev mode](#dev-mode), but they are also minified so they take up less space.
+2. Hashes are generated for asset files like images. These will be used later.
+3. [Webpack](https://webpack.js.org) runs with watch mode disabled and a configuration optimized for application performance.
```
vojtechjelinek, line 68: Maybe link the prod configuration.

Build-process.md
```diff
@@ -0,0 +1,116 @@
...
+Deploy mode is also very similar to [plain prod mode](#plain-prod-mode), except that the `EMULATOR_MODE` constant is set to `false`. Further, we make some changes to the `app.yaml` file:
+
+* Remove the `version: default` line
+* Remove the environment variables specified in the `ENV_VARS_TO_REMOVE_FROM_DEPLOYED_APP_YAML` constant in `build.py`. We remove these since they are already present in the production environment.
+
+This mode is used by our deployment scripts.
+
+#### Unused: Deparallelized terser mode
+
+We used to use a `deparallelize_terser` mode when running the E2E tests on CircleCI, but since we don't run E2E tests on CircleCI anymore, this mode is unused and may be removed soon.
```
vojtechjelinek, line 103: Should we even mention it here? I think it will be removed pretty soon.

vojtechjelinek (changes requested):
Thanks! Left a few comments.

Build-process.md
U8NWXD: done

Build-process.md
U8NWXD: done

Build-process.md
U8NWXD: done

Build-process.md
U8NWXD: done

Build-process.md
U8NWXD: I added a note that maintenance and deploy mode can be enabled together, but I think it still makes sense to discuss them separately because maintenance mode can also be enabled locally (e.g. for testing maintenance mode)

U8NWXD:
@vojtechjelinek PTAL

vojtechjelinek (approved):
LGTM!

DubeySandeep (approved):
@vojtechjelinek Thanks for the review! 

Build-process.md
```diff
@@ -0,0 +1,114 @@
...
+    * [Maintenance mode](#maintenance-mode)
+    * [Deploy mode](#deploy-mode)
+    * [Unused: Deparallelized terser mode](#unused-deparallelized-terser-mode)
+* [Source maps](#source-maps)
+
+## Introduction
+
+Before we can run the Oppia server, we have to build it. This build process transforms the files defining our application (e.g. code, images, configuration files, etc.) from the developer-friendly format we store in GitHub into a form that can be executed to run the server. If you are running the local development server, this build process makes minimal changes, but when deploying to production, we do much more optimization.
+
+This page documents how the build process works in each of its modes. For most developers the build steps are handled automatically and can be ignored; however, some bugs appear only in certain build modes. For example, a bug might only happen in prod mode, in which case a developer will be unable to reproduce it in dev mode.
```
seanlip, line 20: The part of this starting from "however, some bugs appear only in certain build modes" seems unrelated to the part before it. Perhaps you could elaborate a bit more on what you mean by "For most developers the build steps are handled automatically and can be ignored"?

Or, another way to put it is to try and be clear to developers about what they need to do when investigating a bug. How would they know whether the bug appears in dev mode, or prod mode, or both? Should they assume one and, if it's not reproducible there, try the other? etc. -- Some reasoning around the thought process to take would be helpful.

Build-process.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Build script
+
+The build process is performed by [`scripts/build.py`](https://github.com/oppia/oppia/blob/develop/scripts/build.py). If you ever want to know exactly how a particular part of the build process works, you should check out the code there.
+
+## Build modes
+
+There are two primary build modes: dev mode and prod mode. When running in prod mode, you can also enable maintenance mode or deploy mode, or you can choose to only minify third-party libraries. This means the build mode hierarchy looks like this:
```
seanlip, line 28: Are the three prod mode categories mutually exclusive? It's not clear because the third thing isn't a "mode" and sounds like an add-on, but the structure below suggests that they are mutually exclusive things.

EDIT/UPDATE: I saw the sentence after this list. Maybe we should think of these in terms of flags/options rather than modes? The question of which flags are mutually exclusive (if any) still needs to be explained/described, though. It would also be nice to rename things both here and in the codebase so that we don't use "mode" for things that aren't modes.

Build-process.md
```diff
@@ -0,0 +1,114 @@
...
+
+* Dev mode
+* Prod mode
+  * Maintenance mode
+  * Deploy mode
+  * Minify only third-party libraries
+
+Note that maintenance and deploy mode can be enabled simultaneously.
+
+We use prod mode for most everything besides local development. For example, we use prod mode when deploying to production and testing servers, and we run most tests in prod mode. A notable exception is the lighthouse accessibility tests, which run in dev mode because they don't depend on prod mode, and dev mode compilation is faster. Note that we also sometimes use prod mode locally when debugging to run a version of the app that's closer to the deployed version.
```
seanlip, line 38: "which run in dev mode because they don't depend on prod mode" is weird, maybe replace "prod mode" there with "the minifications and optimizations applied in prod mode" or something like that.

seanlip: It might also be a good idea in general to explain the tradeoff between dev / prod mode (less fidelity to what's running in prod vs. longer compilation time). This can help developers have a framework to reason about what approach is best for their use case.

Build-process.md
```diff
@@ -0,0 +1,114 @@
...
+
+
+### Build mode constants
+
+The build mode affects the application during runtimes through the following constants in `assets/constants.ts`:
```
seanlip, line 43: runtimes --> runtime ?

Build-process.md
```diff
@@ -0,0 +1,114 @@
...
+
+* `DEV_MODE`: Set to `true` if and only if we are in dev mode. Otherwise, this is `false`.
+* `EMULATOR_MODE`: Set to `true` if and only if we are not in deploy mode. Note that being in deploy mode implies being in prod mode. Otherwise, this is `false`.
```
seanlip, line 46: "Note that being in deploy mode implies being in prod mode." -- Modes are usually mutually exclusive, so this is kind of confusing and we might want to think about a renaming/reframing here. Ditto below.

Build-process.md
```diff
@@ -0,0 +1,114 @@
...
+* `ENABLE_MAINTENANCE_MODE`: Set to `true` if and only if we are in maintenance mode. Note that being in maintenance mode implies being in prod mode. Otherwise, this is `false`.
+
+### Dev mode
+
+We use dev mode for local development. For example, when you run `python -m scripts.start`, you build the app in dev mode. This mode is optimized to build as quickly as possible, so it doesn't do much optimization.
```
seanlip, line 51: The last sentence is a bit unclear because "optimize" has two different meanings; perhaps clarify or use different wording.

Build-process.md
```diff
@@ -0,0 +1,114 @@
...
+
+Here's what happens during a dev mode build:
+
+1. Some third-party libraries are generated. If you recently ran the server, then there might not be much to change here. Otherwise, the third-party libraries from `manifest.json` are downloaded. Note that the libraries in `manifest.json` are just some of our frontend libraries. The downloaded CSS and JavaScript code files are combined into `third_party/generated/third_party.css` and `third_party/generated/third_party.js`, and the downloaded fonts are placed in `third_party/webfonts/`.
```
seanlip, line 55: What is the sentence "Note that the libraries in `manifest.json` are just some of our frontend libraries." trying to convey? I feel like I don't get it.

Is it related to the stuff that comes after it?

Build-process.md
```diff
@@ -0,0 +1,114 @@
...
+2. The build script sets the constants in `assets/constants.ts` to enable dev mode. Specifically, `DEV_MODE` is `true`, `EMULATOR_MODE` is `true`, and the other constants are `false`.
+3. [Webpack](https://webpack.js.org/) bundles our frontend code files and runs with a [dev configuration](https://github.com/oppia/oppia/blob/develop/webpack.dev.config.ts) designed for minimal compilation time. Note that the dev mode configuration file inherits from the [root webpack config file](https://github.com/oppia/oppia/blob/develop/webpack.common.config.ts). In dev mode, webpack runs in "watch" mode so that when files are changed, webpack re-builds the app automatically. Note that **this step is not handled by `build.py`.** Instead, this step is usually performed by whatever script you actually executed, for example `start.py`. When running the frontend tests, this step is triggered by configuration in `karma.conf.ts`.
```
seanlip, line 57: Maybe worth explaining, conceptually, why this step is not handled by build.py. (Or should it be?)

This is so that devs who edit build.py have a conceptual understanding of what's in scope for that file and what isn't.

Build-process.md
```diff
@@ -0,0 +1,114 @@
...
+### Prod mode
+
+In prod mode, we more aggressively optimize the app for performance, but the cost of this optimization is that the build process takes longer. Prod mode is also a little more complicated because it has three additional options: minify only third-party libraries, maintenance mode, and deploy mode.
+
+#### Plain prod mode
+
+First, let's consider what happens in plain prod mode where those three options are all disabled:
+
+1. The third-party libraries are generated like in [dev mode](#dev-mode), but they are also minified so they take up less space.
+2. Hashes are generated for asset files like images. These will be used later.
```
seanlip, line 69: Maybe change to `later (see step 5 below)`.

Build-process.md
```diff
@@ -0,0 +1,114 @@
...
+Even when running in dev mode, the code your browser runs does not look much like the code on your file system. Files have been combined, and in prod mode, they've been minified. To make debugging easier, you can generate source maps that map from the code your browser is running back to the code on your file system.
+
+Generating source maps is expensive, so by default we don't make them. However, you can enable source map generation in both dev and prod modes, usually by passing a `--source_maps`, for example:
+
+```console
+python -m scripts.start --source_maps
+python -m scripts.run_e2e_tests --prod_env --source_maps
+```
+
+With these flags, the `webpack.*.sourcemap.config.ts` configuration files are used to enable source mapping.
```
seanlip, line 114: Maybe include some explanation on how to use the source maps once they're generated (or link to a page explaining that)? (This also seems like something that should be covered or linked to on a "debugging tips/guide" page.)

seanlip (changes requested):

Build-process.md
U8NWXD: I agree! Since this will require changes to the build code, I opened #13909 to get input from vojtechjelinek

Build-process.md
U8NWXD: I think it should be handled by `build.py`, but I've reached out to @vojtechjelinek to check. If they agree, I'll open an issue

Build-process.md
U8NWXD: done

Build-process.md
U8NWXD: done

Build-process.md
U8NWXD: I'm just trying to make sure I don't give the impression that `manifest.json` is the only place where we specify third-party libraries. I put the sentence in parentheses and added another clause to clarify.

Build-process.md
U8NWXD: done

Build-process.md
U8NWXD: done

Build-process.md
U8NWXD: done

Build-process.md
U8NWXD: I've added a list of all allowed mode combinations to clarify which are mutually exclusive. I agree with renaming the modes, and I've opened #13909 to get @vojtechjelinek's input

Build-process.md
U8NWXD: I re-wrote the second paragraph to be clearer and to point to a new section at the end describing how to think about debugging

U8NWXD:
@seanlip I've addressed most of your comments, except for https://github.com/U8NWXD/oppia-wiki/pull/31#discussion_r714319255 where I'm checking with vojtechjelinek

Build-process.md
vojtechjelinek: If we are talking about FE tests then it is because it is required to be inside karma.

For the start.py it is because the webpack is in watch mode and if this was done in build.py we would have to keep the build part running so that we can keep the watch mode.

U8NWXD: I've updated the page to explain this

U8NWXD:
@seanlip PTAL. I've addressed all comments now

Build-process.md
```diff
@@ -28,45 +29,56 @@ The build process is performed by [`scripts/build.py`](https://github.com/oppia/
...
 
 * Dev mode
+* Minify only mode
 * Prod mode
   * Maintenance mode
   * Deploy mode
-  * Minify only third-party libraries
 
-Note that maintenance and deploy mode can be enabled simultaneously.
+Note that the maintenance and deploy modes can be enabled simultaneously, so you should really think of them as flags. If you're in prod mode, you can also enable maintenance mode dev mode, both, or neither.
```
seanlip, line 37: I don't understand the second sentence.

Build-process.md
```diff
@@ -28,45 +29,56 @@ The build process is performed by [`scripts/build.py`](https://github.com/oppia/
...
 
-We use prod mode for most everything besides local development. For example, we use prod mode when deploying to production and testing servers, and we run most tests in prod mode. A notable exception is the lighthouse accessibility tests, which run in dev mode because they don't depend on prod mode, and dev mode compilation is faster. Note that we also sometimes use prod mode locally when debugging to run a version of the app that's closer to the deployed version.
+Dev mode compiles the fastest, but it produces an application that's least like what we run in production. On the other hand, prod mode takes longer to compile but produces an application that is  the closest to what we run in production. Minify only mode is in the middle.
```
seanlip, line 39: Extra space after "is".

Perhaps, throughout, hyphenate "minify-only" so that "minify" is not read as a verb.

Build-process.md
seanlip: > (This also seems like something that should be covered or linked to on a "debugging tips/guide" page.)

Any thoughts on / plans for this bit?

seanlip (changes requested):
Thanks! Looks better, only minor things.

How do you want to handle the update to modes, documentation-wise? It seems something worth collaborating with @vojtechjelinek on to come up with a scheme that makes sense (where you can represent the developers who are trying to understand the system) but I'm not sure if we should get this PR in and file a follow-up issue, or wait for that to be resolved and then get this PR in.

Build-process.md
U8NWXD: done (both)

Build-process.md
U8NWXD: I re-wrote it to try and be clearer

Build-process.md
U8NWXD: This made me realize that while we have guides for debugging tests, we don't have any guides for debugging the Oppia code directly. This would probably be most useful for debugging the frontend code with the Chrome dev tools. I added #39 to track adding a page for that

U8NWXD:
@seanlip PTAL

> How do you want to handle the update to modes, documentation-wise?

I think we should get this PR in first. Then when @vojtechjelinek and I update the modes in the code, we can update the docs as well.

seanlip (approved):
Sgtm, thanks. Maybe leave an issue to track the doc updates specifically (I don't see it [here](https://github.com/U8NWXD/oppia-wiki/issues)).

U8NWXD:
Done! I opened #40

PR #36, opened 2021-09-23
Update Arch Linux instructions

U8NWXD:


U8NWXD:
@vojtechjelinek @DubeySandeep PTAL. I know you probably can't review for correctness if you don't use Arch Linux, but please just review as best you can

vojtechjelinek:
Actually, I can also test it since I have Arch available on one laptop. Let me know if that would be useful.

U8NWXD:
@vojtechjelinek that would be great since the Arch Linux instructions don't get tested that often

DubeySandeep (approved):

seanlip (approved):
No particular concerns from my end, just noting that I'm not familiar with Arch Linux so my ability to give useful feedback here is limited.

U8NWXD:
@vojtechjelinek are you still able to test these instructions? If so, that would be great

vojtechjelinek:
@U8NWXD Sorry, I was a bit busy with other stuff I will try to look into it in the next two days.

vojtechjelinek (approved):
LGTM! While testing this on my Arch I managed to break my system (but not because of these instructions) so I only tested few thing at the top.

PR #42, opened 2021-09-30
Add pages for debugging custom linter tests

U8NWXD:


Debug-custom-ESLint-check-tests.md
```diff
@@ -0,0 +1,91 @@
...
+* [Print statements](#print-statements)
+* [Inspect the AST](#inspect-the-ast)
+
+## Introduction
+
+Especially for complex lint rules, it can be hard to figure out why a test failed. This page will give you some tips to help you debug your failing tests.
+
+## Print statements
+
+If you add a `console.log()` or `console.error()` statement to your custom ESLint rule code, the output will appear along with test status information. For example, suppose we have a lint rule that raises an error every time `.click()` or `.sendKeys()` is called on an object instead of `action.click()` or `action.sendKeys()`, which we would like people to use instead. If we add `console.log()` statement to print `DEBUGGING found lint error` whenever our lint rule raises an error, we see this output:
```
U8NWXD, line 13:
```suggestion
If you add a `console.log()` or `console.error()` statement to your custom ESLint rule code, the output will appear along with the test status output. For example, suppose we have a lint rule that raises an error every time `.click()` or `.sendKeys()` is called on an object instead of `action.click()` or `action.sendKeys()`, which we would like people to use instead. If we add `console.log()` statement to print `DEBUGGING found lint error` whenever our lint rule raises an error, we see this output:
```

Debug-custom-Pylint-check-tests.md
```diff
@@ -0,0 +1,40 @@
...
+## Introduction
+
+When trying to debug the tests for a custom Pylint checker, remember that the tests are just like any other backend test. You may find our [[guide to debugging backend tests|Debug-backend-tests]] useful. That said, there are some helpful tips for working with Pylint checkers in particular. We describe them below.
+
+## Run a single checker
+
+Debugging a Pylint checker's tests often comes down to debugging the checker itself. To see how your checker behaves on some code, put that code into `test.py`. Then assuming that your checker has message with symbol `my-checker-message`, you can run your checker on `test.py` like this:
+
+```console
+$ PYTHONPATH=scripts/linters ../oppia_tools/pylint-2.8.3/bin/pylint --load-plugins=pylint_extensions --disable=all --enable=my-checker-message test.py
```
U8NWXD, line 17: Note that pylint version may varry

Debug-custom-Pylint-check-tests.md
```diff
@@ -0,0 +1,40 @@
...
+
+```
+
+(These instructions come from the [Pylint documentation](https://pylint.pycqa.org/en/latest/how_tos/custom_checkers.html#debugging-a-checker).)
+
+When running a single checker, two more debugging techniques may be useful: print statements and the Python debugger.
+
+### Print statements
+
+You can print out debugging information from your checker code using `print()` statements. The output will appear when you run the checker as shown above.
```
U8NWXD, line 30: show example

U8NWXD:
@DubeySandeep @Hudda PTAL

Debug-custom-ESLint-check-tests.md
```diff
@@ -0,0 +1,101 @@
...
+* [Print statements](#print-statements)
+* [Inspect the AST](#inspect-the-ast)
+
+## Introduction
+
+Especially for complex lint rules, it can be hard to figure out why a test failed. This page will give you some tips to help you debug your failing tests.
+
+## Print statements
+
+If you add a `console.log()` or `console.error()` statement to your custom ESLint rule code, the output will appear along with the test status output . For example, suppose we have a lint rule that raises an error every time `.click()` or `.sendKeys()` is called on an object instead of `action.click()` or `action.sendKeys()`, which we would like people to use instead. If we add `console.log()` statement to print `DEBUGGING found lint error` whenever our lint rule raises an error, we see this output:
```
Hudda, line 13: With eslint `console.log()` does not print anything on the console. Instead, we have to use `console.info`, even eslint use `console.info` internally to print error logs. So maybe add `console.info` to the list.

Debug-custom-ESLint-check-tests.md
```diff
@@ -0,0 +1,101 @@
...
+      ✓ it('should test a feature', function() {
+        action.sendKeys("Element", elem, "keys");
+      });
+      ✓ it('should test a feature', function() {
+        console.log(elem.click);
+      });
+      ✓ it('should test a feature', function() {
+        console.log(elem.sendKeys);
+      });
+    invalid
```
Hudda, line 30: We don't use this test format for eslint tests, you can see the format for eslint tests here(https://github.com/oppia/oppia/blob/develop/scripts/linters/custom_eslint_checks/rules/sorted-dependencies.spec.js)

U8NWXD: This is copied from the output I got when running `scripts/run_custom_eslint_tests.py`, so I'm pretty sure this is the correct test output format. Note that this is the test *output*, not the test definitions like in the file you linked to

Debug-custom-ESLint-check-tests.md
U8NWXD: done

U8NWXD:
@Hudda @DubeySandeep PTAL

Debug-custom-ESLint-check-tests.md
Hudda: Ah-- I see.

Hudda (approved):

DubeySandeep:

Debug-custom-ESLint-check-tests.md
```diff
@@ -0,0 +1,101 @@
+## Table of contents
+
+* [Introduction](#introduction)
+* [Print statements](#print-statements)
+* [Inspect the AST](#inspect-the-ast)
+
+## Introduction
+
+Especially for complex lint rules, it can be hard to figure out why a test failed. This page will give you some tips to help you debug your failing tests.
```
seanlip, line 9: This intro is a bit sudden, and perhaps more context can be given (e.g. that we're talking specifically about ESlint rules, etc.). Maybe start with something like "Sometimes, you may run into issues when ..." so that the reader can understand whether this page applies to them or not.

Debug-custom-ESLint-check-tests.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Print statements
+
+If you add a `console.log()`, `console.error()`, or `console.info()` statement to your custom ESLint rule code, the output will appear along with the test status output. For example, suppose we have a lint rule that raises an error every time `.click()` or `.sendKeys()` is called on an object instead of `action.click()` or `action.sendKeys()`, which we would like people to use instead. If we add `console.log()` statement to print `DEBUGGING found lint error` whenever our lint rule raises an error, we see this output when running the tests:
```
seanlip, line 13: If we add --> If we add a

Debug-custom-ESLint-check-tests.md
```diff
@@ -0,0 +1,101 @@
...
+DEBUGGING found lint error
+      ✓ it('should test a feature', function() {
+        element(by.css('.protractor-test')).click();
+      });
+      ✓ it('should test a feature', function() {
+        element(by.css('.protractor-test')).sendKeys("keys");
+      });
+```
+
+There are a few important points right away:
```
seanlip, line 49: points --> points to note

Or perhaps better: "However, there are a few issues that you may run into with this approach:" -- that helps prepare the reader better for how the following points fit into the bigger picture.

Debug-custom-ESLint-check-tests.md
seanlip: I'd also recommend not starting with "If you add a ...". Have the "If" describe the scenario the developer is facing and then provide the "add a" as a suggestion of what to do in that scenario.

Otherwise everything seems to be founded on hypotheticals, which is confusing.

Debug-custom-ESLint-check-tests.md
```diff
@@ -0,0 +1,101 @@
...
+
+* All the tests are named `should test a feature`. This is because we generate tests dynamically based on the test cases you provide in your `*.spec.js` file. If you are used to using `fit` to run a single frontend test in isolation, this dynamic test generation means that `fit` won't work.
+* Our debugging statements don't appear in a consistent place with respect to the test status report. For example, the debugging output for the last test appears grouped with the output from the second-to-last test. You shouldn't assume that output from your debugging statements will be synchronized with the test status messages.
+
+These points make debugging a little more difficult, but there's an easy fix: Comment-out all the tests in your `*.spec.js` file except one that you want to debug. Then you'll know that all the debugging statements came from that one test.
```
seanlip, line 54: one --> the one

Debug-custom-ESLint-check-tests.md
```diff
@@ -0,0 +1,101 @@
...
+Considering node starting at line 1
+Considering node starting at line 2
+TypeError: node.callee.object is undefined
+```
+
+This tells us that the error occurs on the second line of the test code. If we copy this test case into AST Explorer, we can see that `foo(1)` is a `CallExpression` but its `callee` attribute has no `object` attribute:
+
+![Screenshot of AST Explorer with test case and AST.](images/debugCustomESLintChecks.png)
+
+This is why we're getting that error! We could fix this problem by checking that `callee` has the `object` and `property` attributes at the start of our rule.
```
seanlip, line 101: Perhaps show the fix, so that it is clear what you are talking about.

Debug-custom-Pylint-check-tests.md
```diff
@@ -0,0 +1,49 @@
...
+  * [Print statements](#print-statements)
+  * [The Python debugger](#the-python-debugger)
+
+## Introduction
+
+When trying to debug the tests for a custom Pylint checker, remember that the tests are just like any other backend test. You may find our [[guide to debugging backend tests|Debug-backend-tests]] useful. That said, there are some helpful tips for working with Pylint checkers in particular. We describe them below.
+
+## Run a single checker
+
+Debugging a Pylint checker's tests often comes down to debugging the checker itself. To see how your checker behaves on some code, put that code into `test.py`. Then assuming that your checker has message with symbol `my-checker-message`, you can run your checker on `test.py` like this:
```
seanlip, line 14: Perhaps describe what sort of output is expected to be produced by this process, so that it's clear to the reader how they can use that output for debugging.

seanlip (changes requested):

PR #43, opened 2021-09-30
Add 'Writing Apache Beam Jobs' section to wiki

brianrodri:


Apache-Beam-Jobs.md
```diff
@@ -227,6 +228,82 @@ error_pcoll = (
 
 `Runner`s provide the `run()` method used to visit every node (`PValue`) in the pipeline's DAG by executing the edges (`PTransform`s) to compute their values.  At Oppia, we use `DataflowRunner` to have our `Pipeline`s run on the [Google Cloud Dataflow service](https://cloud.google.com/dataflow).
 
+## Writing Apache Beam Jobs
```
U8NWXD, line 231: I feel like this section needs more setup so that a contributor who is new to beam jobs doesn't get lost

Apache-Beam-Jobs.md
```diff
@@ -227,6 +228,82 @@ error_pcoll = (
...
+
+* In English, this means that **the job _must_ report _something_ about what occurred during its execution.** For example, this can be the errors it discovered or the number of successful operations it was able to perform. **Empty results are forbidden!**
+
+  * If you don't think your job has any results worth reporting, then just print a "success" metric with the number of models it processed.
+
+* `JobRunResult` outputs should answer the following questions:
+
+  * Did the job run without any problems? How and why do I know?
+  * How much work did the job manage to do?
+  * If the job encountered a problem, what caused it?
```
U8NWXD, line 287: Should we remove the duplicate of this content in the case study below?

Apache-Beam-Jobs.md
```diff
@@ -227,6 +228,82 @@ error_pcoll = (
...
+* `dashboard_stats_computation_jobs.py`
+* `exploration_indexing_jobs.py`
+* `exploration_stats_regeneration_jobs.py`
+* `model_validation_jobs.py`
+
+**New modules must be imported in the `jobs/registry.py` file.**
+
+---
+
+The `run()` method must return a `PCollection[JobRunResult]`.
```
U8NWXD, line 277: I think it would be clearer to say "PCollection of JobRunResults" since not everyone is going to be familiar with the `PCollection[JobRunResult]` syntax

Apache-Beam-Jobs.md
```diff
@@ -227,6 +228,82 @@ error_pcoll = (
...
+  * How much work did the job manage to do?
+  * If the job encountered a problem, what caused it?
+
+When implementing the `run()` method, make liberal usage of small and simple `PTransform`s and `DoFn`s. You must also be careful to only use the following legal constructs:
+
+| Operation       | Do use                                 | Do not use                     |
+| --------------- | -------------------------------------- | ------------------------------ |
+| Getting models  | ndb_io.GetModels(ModelName.query(...)) | ModelName.get_multi()          |
+| Putting models  | model_pcoll \| ndb_io.PutMulti()       | ModelName.put_multi(models)    |
+| Deleting models | model_pcoll \| ndb_io.DeleteMulti()    | ModelName.delete_multi(models) |
```
U8NWXD, line 295: These are confusing because the "Do use" items haven't been introduced yet

Apache-Beam-Jobs.md
```diff
@@ -227,6 +228,82 @@ error_pcoll = (
...
+
+  * If you don't think your job has any results worth reporting, then just print a "success" metric with the number of models it processed.
+
+* `JobRunResult` outputs should answer the following questions:
+
+  * Did the job run without any problems? How and why do I know?
+  * How much work did the job manage to do?
+  * If the job encountered a problem, what caused it?
+
+When implementing the `run()` method, make liberal usage of small and simple `PTransform`s and `DoFn`s. You must also be careful to only use the following legal constructs:
```
U8NWXD, line 289: > You must also be careful to only use the following legal constructs:

This makes it sound like the items in the table below are the only allowed PTransforms in jobs. Do you mean that these are the only allowed ways to get, put, and delete models?

Apache-Beam-Jobs.md
```diff
@@ -227,6 +228,82 @@ error_pcoll = (
...
+
+If you think you need to use something else -- e.g. files/etc. -- please talk to @brianrodri/@vojtechjelinek first. Some of these -- typically operations defined by datastore_services -- don't work on Apache Beam in production, while they might work locally.
+
+---
+
+When testing jobs, always inherit from `JobTestBase` and override the class constant `JOB_CLASS`. You can then run `self.assert_job_output_is(...)` or `self.assert_job_output_is_empty()` to verify behavior.
+
+When testing `PTransform`s and `DoFn`s, always inherit from `PipelinedTestBase`.  You can use `self.assert_pcoll_is(...)` or `self.assert_pcoll_is_empty(...)` to verify behavior.
+
+* **IMPORTANT:** Only one `assert_(job|pcoll)_is` assertion can be performed in a test body. Multiple calls will result in an exception instructing you to split the test apart.
```
U8NWXD, line 305: I think it might be worth fleshing out this section some more to describe good testing practices. Maybe provide examples? Also, we should probably specify whether these are written like usual backend tests and, if so, link to the backend tests wiki page

U8NWXD (changes requested):

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
brianrodri: How's this?

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
```diff
@@ -227,6 +229,257 @@ error_pcoll = (
...
+
+    .--------------. Count states .--------. Sum .-------.
+    | Explorations | -----------> | Counts | --> | Total |
+    '--------------'              '--------'     '-------'
+
+### 1. Subclass the `base_jobs.JobBase` class and override the `run()` method
+
+Make sure your job class name is clear and concise, because the name is presented to release coordinators:
+
+<!-- TODO(#11475): Replace this screenshot with one that has job names which follow our naming convention. -->
```
seanlip, line 246: I don't see this tracked in #11475.

It's also on a separate repo so maybe better to just fix it (through in-browser html manipulation) and get rid of the todo?

Apache-Beam-Jobs.md
```diff
@@ -227,6 +229,257 @@ error_pcoll = (
 
 `Runner`s provide the `run()` method used to visit every node (`PValue`) in the pipeline's DAG by executing the edges (`PTransform`s) to compute their values.  At Oppia, we use `DataflowRunner` to have our `Pipeline`s run on the [Google Cloud Dataflow service](https://cloud.google.com/dataflow).
 
+## Writing Apache Beam Jobs
+
+For this section, we'll walk through the steps of implementing a job by writing one: `CountExplorationStatesJob`.
```
seanlip, line 234: Explain what your example job does, so that the contributor knows what they're trying to do and what the problem statement actually is. There should probably be some planning stage at the outset for "how" to do it / how to think through the impl, maybe you can guide them through the general thinking that would be needed as step 0.

Apache-Beam-Jobs.md
```diff
@@ -227,6 +229,257 @@ error_pcoll = (
...
+    * `exploration_stats_regeneration_jobs.py`
+    * `model_validation_jobs.py`
+
+    However, you should always prefer placing jobs in pre-existing modules if an appropriate one already exists.
+
+For this example, we will write our job in the module: `core/jobs/batch_jobs/exploration_inspection_jobs.py`.
+
+### 2. Override the `run()` method to operate on `self.pipeline`
+
+As illustrated in the Architecture section, jobs are organized by `Pipeline`s, `PTransform`s, and `PCollection`s. Thus, when writing your job, **you must build _off_ of your pipeline** in order to construct the DAG. We can represent this by adding a special `Pipeline` node signifying the _true_ beginning of the DAG:
```
seanlip, line 290: I don't really know what "my pipeline" is. Can you explain conceptually what's happening here and why the pipeline node is the true beginning of the DAG? Does my pipeline already exist or is it a new thing that I am constructing as part of implementing the job? (The current doc suggests the former.)

Apache-Beam-Jobs.md
```diff
@@ -227,6 +229,257 @@ error_pcoll = (
...
+                exp_models.ExplorationModel.get_all(deleted=False))
+        )
+```
+
+Observe that:
+1. We're using `self.pipeline` as if it were a `PValue`
+2. We're using `ndb_io.GetModels` rather than `get_multi`
+3. We're passing a Query to `ndb_io.GetModels`
+
+It just so happens that `Pipeline` does indeed count as a `PValue`, with the special interface "`PBegin`". `PBegin` values signify that the value is never the output of a `PTransform`. Operations can branch off of `PBegin` values, like `ndb_io.GetModels()` does, to begin execution on a pipeline.
```
seanlip, line 328: "It just so happens" suggests coincidence/chance. Instead, try to explain how readers can think of Pipelines as PValues conceptually -- e.g. "Note that a `Pipeline` is a special case of a `PValue` that ..."

Apache-Beam-Jobs.md
```diff
@@ -227,6 +229,257 @@ error_pcoll = (
...
+| `delete_multi(keys)`   | `key_pcoll \| ndb_io.DeleteMulti()`          |
+
+Note that `get_multi` has the biggest change in interface, in that it takes a `Query` argument. You can get a query for any model by using the class method `get_all`.
+-   **IMPORTANT:** Never use `datastore_services.query_everything()`!! Due to a limitation in Apache Beam, this operation is incredibly slow and inefficient! **You are almost certainly doing something wrong if you need this function.** Ask @brianrodri/@vojtechjelinek for help if you believe you need to use it regardless.
+
+Why should we use these `PTransform` over the simpler `get`/`put`/`delete` functions? **Performance**. The `get`/`put`/`delete` function calls are all *synchronous*, so your job's performance will suffer greatly by waiting for the operations to complete.
+
+`PTransform`s, on the other hand, are specially crafted to take advantage of the Apache Beam framework and guarantee better performance. In general, **you should always prefer `ndb_io` over any of the `get`/`put`/`delete` functions!** If you think you have a valid need for avoiding `ndb_io`, then speak with @brianrodri/@vojtechjelinek first.
```
seanlip, line 343: Great caveats :)

Apache-Beam-Jobs.md
```diff
@@ -227,6 +229,257 @@ error_pcoll = (
...
+            exp_models.ExplorationModel.get_all(deleted=False))
+    )
+
+    state_count_pcoll = (
+        exp_model_pcoll
+        | 'Count states' >> beam.Map(self.get_number_of_states)
+    )
+```
+
+Note that we chose to use `beam.Map` here instead of `beam.ParDo`. This is mostly a stylistic choice, as `beam.Map` is just a specialized version of `ParDo`, in that `Map` simply takes each input element and "maps" them to a single output element. In our case, each `ExplorationModel` will map to a single `int`, the number of states.
```
seanlip, line 361: them --> it

Apache-Beam-Jobs.md
```diff
@@ -227,6 +229,257 @@ error_pcoll = (
...
+
+```python
```
seanlip, line 363: Introduce this: "Here is the implementation of get_number_of_states. This function just transforms the model into ..." (and you can drop the one-sentence paragraph after the code snippet).

Apache-Beam-Jobs.md
```diff
@@ -227,6 +229,257 @@ error_pcoll = (
...
+def get_number_of_states(self, exp_model: ExplorationModel) -> int:
+    exp = exp_fetchers.get_exploration_from_model(exp_model)
+    return len(exp.states)
+```
+
+Our implementation is simple, we transform the model into a domain object and measure the length of its `states` dictionary.
+
+Finally, we need to sum all the counts together. Fortunately, Apache Beam already has a `PTransform` that does exactly this:
```
seanlip, line 371: I don't immediately see what this PTransform in question is when looking at the code. Maybe move the 'CombineGlobally' part of the explanation up here.

Apache-Beam-Jobs.md
```diff
@@ -227,6 +229,257 @@ error_pcoll = (
...
+
+  * If you don't think your job has any results worth reporting, then just print a "success" metric with the number of models it processed.
+
+* `JobRunResult` outputs should answer the following questions:
+
+  * Did the job run without any problems? How and why do I know?
+  * How much work did the job manage to do?
+  * If the job encountered a problem, what caused it?
+
+Our job is trying to report the total number of states across all explorations, so we need to create `JobRunResult` that holds that information. For this, we can simply use the `as_stdout` helper method:
```
seanlip, line 408: create --> create a

Apache-Beam-Jobs.md
```diff
@@ -227,6 +229,257 @@ error_pcoll = (
...
+        'e1', 'o1', ['I', 'J'], ['TextInput'])
+    self.save_new_linear_exp_with_state_names_and_interactions(
+        'e1', 'o1', ['K', 'L', 'M', 'N'], ['TextInput'])
+
+    self.assert_job_output_is([
+        job_run_result.JobRunResult(stdout='14'),
+    ])
+```
+
+Note that `self.assert_job_output_is(...)` and `self.assert_job_output_is_empty()` do as advertised, it runs the job to completion and verifies the result.
```
seanlip, line 479: ... as advertised -- they run the job to completion and verify the result.

Apache-Beam-Jobs.md
```diff
@@ -227,6 +229,257 @@ error_pcoll = (
...
+
+The current convention is to name your test cases `<JobName>Tests`, but you can create better names if you want to break tests up into groups of specialized focus. For our example, we'll keep things simple.
+
+```python
+class CountExplorationStatesJobTests(test_jobs.JobTestBase):
+
+    JOB_CLASS = CountExplorationStatesJob
+```
+
+### 2. Run assertions using a `assert_job_output_is_*` method
```
seanlip, line 449: Explain best practices for figuring out which tests to write.

Explain what sorts of things can and cannot be tested. How will we verify the parts that cannot be tested through this approach?

seanlip (changes requested):

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
brianrodri: Ack'd!

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
seanlip: Hm. How about the second part? 

> Explain what sorts of things can and cannot be tested. How will we verify the parts that cannot be tested through this approach?

What do these tests guarantee -- will they guarantee a successful and correct prod run? If not, what else is needed?

seanlip (changes requested):
Thanks @brianrodri! Just one more note from me.

Apache-Beam-Jobs.md
brianrodri: How's this?

Apache-Beam-Jobs.md
```diff
@@ -433,12 +419,52 @@ def run(self):
...
 
-    return state_count_sum_pcoll | beam.Map(job_run_result.JobRunResult.as_stdout)
+    return (
+        state_count_sum_pcoll
+        | 'Map as stdout' >> beam.Map(job_run_result.JobRunResult.as_stdout)
+    )
 ```
 
-The method maps every element in a `PCollection` into a `JobRunResult` with the stringified-value as its `stdout`.
+The method maps every element in a `PCollection` to a `JobRunResult` with their stringified-values as `stdout`.
```
seanlip, line 428: Missing "its" before "stdout", perhaps? (I'm not really sure what the sentence as-written means.)

Apache-Beam-Jobs.md
```diff
@@ -491,9 +517,24 @@ def test_many_explorations(self):
 
 Note that `self.assert_job_output_is(...)` and `self.assert_job_output_is_empty()` do as advertised -- they run the job to completion and verify the result.
 
--   **IMPORTANT:** Only one `assert_job_output_is` assertion can be performed in a test body. Multiple calls will result in an exception instructing you to split the test apart.
+> **IMPORTANT:** Only one `assert_job_output_is` assertion can be performed in a test body. Multiple calls will result in an exception instructing you to split the test apart.
+
+Just because a job passes in unit tests does not guarantee it will pass in production. This is because workers, which execute the pipeline code, are run in a special environment where the code base is configured differently. While the Apache Beam jobs team works to cut down on the differences, be careful about using complex and/or confusing objects. The simpler your job, the greater chance it'll work in production!
```
seanlip, line 522: Maybe say who the "Apache Beam jobs team" is,  otherwise this seems like the Apache Beam project's maintainers.

seanlip: Can you also explain or give examples of what sorts of things might fail in this way?

seanlip (changes requested):

Apache-Beam-Jobs.md
brianrodri: Hmm I don't think it's something worth sharing, they're tiny environmental differences that won't mean anything to job writers so I don't think there's any value in listing examples.

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
```diff
@@ -1,51 +1,39 @@
...
+        -   [`Map` and `FlatMap`](#map-and-flatmap)
+        -   [`Filter`](#filter)
+        -   [`GroupByKey`](#groupbykey)
+            -   [Example of using `GroupByKey`,`Filter`, and `FlatMap`](#example-of-using-groupbykeyfilter-and-flatmap)
+    -   [`Runner`s](#runners)
+-   [Writing Apache Beam Jobs](#writing-apache-beam-jobs)
+-   [Testing Apache Beam Jobs](#testing-apache-beam-jobs)
+-   [Running Apache Beam Jobs](#running-apache-beam-jobs)
+-   [Case studies](#case-studies)
+    -   [Case Study: `SchemaMigrationJob`](#case-study-schemamigrationjob)
```
U8NWXD, line 18: Can you preserve the original format (e.g. with asterisks, not so many spaces after each bullet)? The original table of contents was auto-generated

Apache-Beam-Jobs.md
```diff
@@ -227,158 +215,386 @@ error_pcoll = (
...
+    '--------------'              '--------'     '-------'
 
-### Case study: `CountAllModelsJob`
+> **TIP**: As illustrated, you don't need to know what the names of the `PTransform`s (edges) used in a diagram are. It's easy to look up the appropriate `PTransform` after drawing the diagram.
 
-**Difficulty:** Trivial
+Now that we have our bearings, let's get started on implementing the job.
 
-**Key Concepts:**
+### 1. Subclass the `base_jobs.JobBase` class and override the `run()` method
```
U8NWXD, line 234: Can you update the table of contents to include this step? We've included other level-3 headings in the TOC

Apache-Beam-Jobs.md
```diff
@@ -227,158 +215,386 @@ error_pcoll = (
...
 
-* Fetching NDB models
-* Counting elements in a `PCollection`
-* Creating `JobRunResult` values
-* Job registration
+Make sure your job class name is clear and concise, because the name is presented to release coordinators:
 
----
+![image](https://user-images.githubusercontent.com/5094060/135734501-eb9c0370-e98d-4271-b41a-0bf11c25503c.png)
```
U8NWXD, line 238: I was thinking about something more descriptive than "image." Could the alt text more clearly explain what the image shows?

U8NWXD (changes requested):

Apache-Beam-Jobs.md
seanlip: Let me put it another way -- what are the steps that are needed to guarantee that things will work in prod? What instructions should the developer give the release team / testing team? Are there certain types of jobs that are known to definitely work?

We can't really have the flow be "write a job and it may or may not work properly" ... we would need some sort of process to introduce new jobs into the codebase, and this doc should be clear on what that process is. Maybe we can discuss tomorrow when meeting with @vojtechjelinek.

Apache-Beam-Jobs.md
brianrodri: Whoops, this was more descriptive but it got deleted. Fixed!

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
brianrodri: As discussed offline, this is not valuable to job writers. Instead, I've added a section instructing job writers to have their jobs tested on production, and linking to the appropriate details page.

seanlip (approved):
Thanks, LGTM!

Apache-Beam-Jobs.md
```diff
@@ -227,107 +223,357 @@ error_pcoll = (
...
+
+> **IMPORTANT:** Only one `assert_job_output_is` assertion can be performed in a test body. Multiple calls will result in an exception instructing you to split the test apart.
+
+Just because a job passes in unit tests does not guarantee it will pass in production. This is because workers, which execute the pipeline code, are run in a special environment where the code base is configured differently. While Oppia's jobs team works to resolve the differences, be careful about using complex and/or confusing objects. The simpler your job, the greater chance it'll work in production!
+
+## Running Apache Beam Jobs
+
+### Local Development Server
+
+These instructions assume you are running a local development server. If you are a release coordinator running these jobs on the production or testing servers, you should already have been granted the "Release Coordinator" role, so you can skip steps 1-3.
```
vojtechjelinek, line 538: Should we mention somewhere that Beam validates Python types and that they shouldn't be left as the last step?

Apache-Beam-Jobs.md
```diff
@@ -227,107 +223,357 @@ error_pcoll = (
...
+![Screen recording showing how to run jobs](https://user-images.githubusercontent.com/5094060/128743997-70cca5f9-0b76-4294-806e-f65f5df5be95.gif)
+
+### Production Server
+
+Before a job or feature that cannot be fully tested locally can be run and deployed in production, it must first be tested on the Oppia backup server. Examples of features that require backup server testing are features that add or modify code that requires some third-party API, like Cloud Tasks, Cloud Storage, or Cloud Translate.
+
+If your job or feature is not essential for the release and has not been fully tested by the release cut, then it is not going into the release. "Fully tested" means:
+- The job or feature should run without failures on the Oppia backup server.
+- The job or feature produces the expected output.
+- The job or feature has the expected outcome (this must be verified by e.g. user-facing changes, or introspection of the datastore, or a validation job, or an output check, etc.).
```
vojtechjelinek, line 556: We should not use datastore introspection except for rare cases where it is the only possible solution.

Apache-Beam-Jobs.md
```diff
@@ -227,107 +223,357 @@ error_pcoll = (
...
+- The job or feature should be explicitly approved by server jobs admin (currently @seanlip and @vojtechjelinek).
+
+Also, in case your job is a new migration job (not any already existing migrations), there has to be an audit job accompanying it to verify that the data that you are migrating is valid in the server. **The audit job will have to be "Fully tested" before testing on the migration job can start.**
```
vojtechjelinek, line 559:
```suggestion
Also, in case your job changes data in the datastore, there has to be a validation job accompanying it to verify that the data that you are migrating is valid in the server. **The validation job will have to be "Fully tested" before testing on the migration job can start.**
```

vojtechjelinek (changes requested):
Thanks! Left a few comments.

Apache-Beam-Jobs.md
brianrodri: Done.

Apache-Beam-Jobs.md
brianrodri: OK, I've omitted it.

Apache-Beam-Jobs.md
brianrodri: Done.

U8NWXD (approved):

vojtechjelinek (approved):
LGTM!

